PR #4, opened 2015-10-07
Model

andy-faunadb:


eaceaser:
first pass looks good. i'll take a more detailed look again in a bit

src/model/Builtin.js
```diff
@@ -0,0 +1,132 @@
...
+  static async createForModel(client, modelClass, data={}) {
+    return await this.create(client, Object.assign({name: modelClass.faunaClassName}, data))
+  }
+
+  /** Gets the class associated with a {@link Model}. */
+  static async getForModel(client, modelClass) {
+    try {
+      await this.get(client, modelClass.classRef)
+    } catch (err) {
+      console.log(err)
```
eaceaser: probably should fail the promise here

src/model/Model.js
```diff
@@ -0,0 +1,347 @@
...
+          setPath(field.path, value, this._current)
+        }
+      } : {
+        get() {
+          if (fieldName in this._cache)
+            return this._cache[fieldName]
+          else {
+            const encoded = getPath(field.path, this._current)
+            const decoded = field.codec.decode(encoded, this)
+            this._cache[fieldName] = decoded
```
eaceaser: are codecs expected to be expensive? If not, I'd say that we do away with the cache entirely and call the codec on get/set unconditionally

src/model/Model.js
```diff
@@ -0,0 +1,347 @@
...
+        throw new InvalidValue(`No such field ${fieldName}`)
+      // This calls the field's setter.
+      this[fieldName] = data[fieldName]
+    }
+  }
+
+  /** {@link Ref} of this instance in the database. Fails if {@link isNewInstance}. */
+  get ref() {
+    if (this.isNewInstance())
+      throw new InvalidQuery('Instance has not been saved to the database, so no ref exists.')
```
eaceaser: I feel like getters should return null, rather than throw, in this case? I kind of feel like its a bit weird for getters to throw, but maybe its not that weird.

src/model/Builtin.js
andy-faunadb: Oops, debug code got left in. Usually I remember to leave a comment so the linter will remind me to delete these.

src/objects.js
```diff
@@ -130,6 +146,70 @@ export class Page {
...
+
+    this.direction = undefined
+    this.cursor = undefined
+  }
+
+  /**
+   * Moves this iterator to the next page and returns its data.
+   * May set `this.done` to `true`.
+   */
+  async next() {
```
eaceaser: I'm not 100% sold on the current PageIterator API. 

It's currently in this weird half-async world, where the deferred methods end up manipulating the object's state, which will then be inspected synchronously. That seems weird to me, and, without async/await, I think the call sites will end up having to jump through hoops (recursion within a promise flatMap) in order to bridge the gap between the Promise monad and the iterative approach that having a `.done` boolean implies.

Without higher level constructs such as async/await, I think the API needs to be purely functional. In the end, it should probably behave more like a stream of pages, rather than an iterator; you should be able to apply a predicate to the stream (i.e. a `.takeWhile()`) method, and then you can act on the stream through a `.foreach` or a `.map`.

Let me know what you think.

andy-faunadb: Is there an existing interface for async iterators/streams that you know of?

eaceaser: I'm not sure about the JS world, but modeling something simple after Scala's http://www.scala-lang.org/api/current/index.html#scala.collection.immutable.Stream probably would be fine

src/model/Model.js
andy-faunadb: I think it's fine. Better than silently handing the user bad data.

freels: What is the js convention? In ruby this would just hand back nil, and the user would be expected to deal with it.

eaceaser: pretty sure its just return null, at least in my experience

eaceaser:
Sweet. this interface is dope. In theory, once the world gets async/await in 100 years, alot of this stuff could be left up to the caller to implement. But I think this maps really well to how people will be dealing with pages in the real world.

eaceaser:
ship this shit

PR #6, opened 2015-10-09
Clean up Client

andy-faunadb:


eaceaser:
lgtm. as an aside, is there a configuration package out there that can handle the config-related object logic? like deafults and whatnot?

andy-faunadb:
Do you mean, is there a package out there implementing `applyDefaults`? Possibly, but I don't know how to find it.

eaceaser:
i mean, this sort of behavior would be common to any library out there that handles configuration. Maybe this? https://github.com/lorenwest/node-config

PR #9, opened 2015-10-21
Write README

andy-faunadb:


README.md
```diff
@@ -1,11 +1,100 @@
...
+    async function testPing() {
+      console.log(await rootClient.ping())
+    }
+
+    async function createDBAndGetClient() {
+      const dbName = 'prydain'
+      const dbRef = new Ref('databases', 'prydain')
+
+      try {
+        await rootClient.delete(dbRef)
```
eaceaser: I don't think the example should use async/await. Most users will not have access to these constructs.

andy-faunadb: I have an ES5 version, use that alone then? (no 'import', 'async', arrow functions)

eaceaser: I think an ES5 version there would be good. Having both couldn't hurt I guess.

eaceaser:
other than having an es5 example everything looks good

PR #11, opened 2015-10-22
Make error handling for bad class work better

andy-faunadb:


src/model/Model.js
```diff
@@ -268,7 +268,7 @@ export default class Model {
     if (!(typeof resource === 'object' && resource.constructor === Object))
       throw new Error('Expected to initialize from plain object resource.')
     const expectedClass = this.constructor.classRef
-    if (!resource.class.equals(expectedClass))
+    if (!(resource.class instanceof Ref && resource.class.equals(expectedClass)))
```
eaceaser: i'd rewrite this as:

`!(resource.class instanceof Ref) || !resource.class.equals(expectedClass)`

Having to demorgan the negation makes it less clear what the intended logic is.

andy-faunadb: This is why we need `unless`

eaceaser: yuck! even worse, lol

eaceaser:
lgtm

PR #16, opened 2015-10-23
Revise parameters of client methods.

andy-faunadb:
* delete does not take a 'data' param
* ping's parameters are optional


eaceaser:
lgtm

PR #21, opened 2015-10-30
Add `toLambda`

andy-faunadb:


test/query_test.js
```diff
@@ -78,19 +78,17 @@ describe('query', () => {
...
     const page = query.paginate(nSet(1))
-    const ns = query.map(getN, page)
+    const ns = query.map(a => query.select(['data', 'n'], query.get(a)), page)
     assert.deepEqual((await client.query(ns)).data, [1, 1])
   })
 
   it('foreach', async function() {
     const refs = [(await create()).ref, (await create()).ref]
     await client.query(
-      query.foreach(query.lambda(query.delete_expr), refs))
```
eaceaser: since this still technically supports passing in a lambda construct as opposed to a function, (query.lambda, etc) we should probably have a test for that still too. unless there is one somewhere outside of the diff and im missing it

andy-faunadb: I added a test of query.map that uses manual lambda.

eaceaser:
ship ti

PR #23, opened 2015-11-02
Move lambda arguments to the end to match faunadb-ruby

andy-faunadb:
This is a breaking change, so we might need to tell the client about this / change the version number.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "faunadb",
-  "version": "0.0.0",
+  "version": "0.1.0",
   "description": "FaunaDB (faunadb.com) client for node",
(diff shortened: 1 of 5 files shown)


evan:
Changing the version should be ok; we should minimize breaking changes when practical, but if we are at a pre-1.0 version number pretty much any release is "allowed" to be breaking.

eaceaser:
can we just use semver-style shit?

just start at 0.1.0, and increment the minor for any breaking change, until we finally release 1.0

andy-faunadb:
We're not changing faunadb-python to match faunadb-ruby, but still, I think putting a function at the end is standard for js libraries.

eaceaser:
lgtm

PR #24, opened 2015-11-03
Clients are now called drivers

andy-faunadb:


andy-faunadb:
Summary of chat: since we have a Client class, it would be confusing to call the project a "driver".

PR #27, opened 2015-11-05
Fixed class permissions test (can't set to "public" any more)

andy-faunadb:


freels:
"public" is still a valid permission, but they have always applied to the individual permissions: "read", "write", "create".

andy-faunadb:
OK. Reading http://localhost:3000/documentation#permissions it seems a bit ambiguous what "permissions themselves" mean (the value of the "permissions" field VS one of the values in the "permissions" object). I like the test a bit better when it involves using a ref so I think I'll leave it.

eaceaser:
ship

PR #28, opened 2015-11-05
Cleanup handling of varargs

andy-faunadb:
`varargsQuery` was unnecessarily complicated and wouldn't work for
functions that had other arguments in addition to the rest arg.


eaceaser:
this looks good. one question though: do you think its worth asserting that the first element of a single-term varargs IS an array?

andy-faunadb:
Technically, `query.equals(1)` is a valid query returning true. There are other examples, all of them similarly silly; there's no real reason to pass in a single argument, but I think throwing an error harms more than it helps. What if someone writes `query.equals(...args)` and it happens that `args.length === 1`?

eaceaser:
word. ship it

PR #32, opened 2015-11-09
Since the install script breaks in node 5.0.0, include `lib` in the repo instead

andy-faunadb:
Fixes https://github.com/faunadb/faunadb-js/issues/30


eaceaser:
ok this looks good

PR #33, opened 2015-11-09
Added index.js and used it in examples

### LINKED ISSUE #31: [bug] `require('faunadb')` or `import faunadb from 'faunadb'` throws an error
The following error is being thrown when doing `import faunadb from "faunadb"`

```
Error: Cannot find module 'faunadb'
    at Function.Module._resolveFilename (module.js:336:15)
    at Function.Module._load (module.js:286:25)
    at Module.require (module.js:365:17)
    at require (module.js:384:17)
    at Object.<anonymous> (/Users/dremonkey/Projects/Javascript/skystream/app/index.js:5:36)
    at Module._compile (module.js:434:26)
    at normalLoader (/Users/dremonkey/Projects/Javascript/skystream/node_modules/babel/node_modules/babel-core/lib/api/register/node.js:199:5)
    at Object.require.extensions.(anonymous function) [as .js] (/Users/dremonkey/Projects/Javascript/skystream/node_modules/babel/node_modules/babel-core/lib/api/register/node.js:216:7)
    at Module.load (module.js:355:32)
    at Function.Module._load (module.js:310:12)
```

This is caused because there is no `index.js` file at the toplevel of the faunadb directory or `main` entry in package.json. From looking at your example I assume you already know this, but for somebody new to the library it isn't clear how exactly one should go about using it.

There are three possible solutions to this... just going to present them so you can choose which you want to go with.

1. Create a top level `index.js` file and use it to expose at the very least the Client class. I'm not sure what else you would like to expose.... `Model` maybe?
2. Add the `main` property to `package.json` that points at one of the files... maybe `lib/Client` if that is all you want to expose or create a `lib/index.js` file and point it there.
3. Mention in the README that in order to actually use `Client` or `Model` or any of the others you have to import it directly `var Client = require("faunadb/lib/Client")` or `import Client from "faunadb/lib/Client")`

The first or second option is probably the method I've seen used most in other node libraries... implementation becomes pretty straightforward too...

```
import faunadb as "faunadb";

let Client = faunadb.Client;
let Model = faunadb.Model;
```

andy-faunadb:
Fixes #31.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,0 +1,22 @@
+/* global exports */
+/* eslint-disable no-var */
+
+var singleExports =
+  ['model/Codec', 'model/Field', 'model/Model', 'AsyncStream', 'Client', 'PageStream']
+singleExports.forEach(function(filename) {
+  var module = require('./lib/' + filename)
+  exports[module.name] = module
+})
+
+var multipleExports = ['model/Builtin', 'errors', 'objects']
+multipleExports.forEach(function(filename) {
+  var module = require('./lib/' + filename)
+  if ('default' in module) {
+    exports[module.default.name] = module.default
+    delete module.default
+  }
+  for (var key in module)
+    exports[key] = module[key]
+})
+
+exports.query = require('./lib/query')
--- a/examples/faunadb-test-es7.js
+++ b/examples/faunadb-test-es7.js
@@ -5,12 +5,7 @@ To run this:
   npm run faunadb-test-es7
 */
 
-import Client from 'faunadb/lib/Client'
-import {NotFound} from 'faunadb/lib/errors'
-import Model from 'faunadb/lib/model/Model'
-import {Class, Database, Key} from 'faunadb/lib/model/Builtin'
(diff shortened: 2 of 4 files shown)


andy-faunadb:
Think I should change tests to use index too? This would make them use `lib`, so you'd have to run `npm run watch` while testing code to keep `lib` up-to-date.

index.js
```diff
@@ -0,0 +1,36 @@
...
+      delete module.default
+      module[dfault.name] = dfault
+    }
+    return module
+  } else {
+    var obj = {}
+    obj[module.name] = module
+    return obj
+  }
+}))
```
eaceaser: I'd prefer that we just set the exports manually rather than through this helper. There's not very many, and its much clearer to do it by hand than to try to infer the shape of the underlying module.

eaceaser:
hmmmmm wrt the tests stuff, i dont think so. I think that the tests should be runnable without an intermediate step.

eaceaser:
ship

PR #37, opened 2015-11-12
Add @ts and @date decode/encode and date/time functions

andy-faunadb:


src/objects.js
```diff
@@ -147,3 +147,61 @@ export class Page {
...
 }
+
+/** FaunaDB time. See the [docs](https://faunadb.com/documentation/queries#values-special_types). */
+export class FaunaTime {
+  /**
+   * @param {string|Date} value If a Date, this is converted to a string.
+   */
+  constructor(value) {
+    if (value instanceof Date)
+      value = value.toISOString()
```
eaceaser: i seem to recall a conversation about this somewhere. In the Java/scala client, I store the native date/time types, and then convert them to string only at serialization time. Should we do that here?

test/objects_test.js
```diff
@@ -49,4 +49,39 @@ describe('objects', () => {
...
+    assert.deepEqual(new FaunaTime(dt).date, dt)
+
+    const epoch = new Date(Date.UTC(1970, 0, 1))
+    const ft = new FaunaTime(epoch)
+    assert.deepEqual(ft, new FaunaTime('1970-01-01T00:00:00.000Z'))
+    assert.deepEqual(ft.date, epoch)
+  })
+
+  it('time', () => {
+      const test_ts = new FaunaTime('1970-01-01T00:00:00.123456789Z')
```
eaceaser: in theory the @ts string can contain a TZ offset, although its going to get normalized by the server.

src/objects.js
andy-faunadb: The native Date (actually a time) type doesn't have nanosecond precision, so we can't use it for FaunaTime.
We could use Date for FaunaDate, although we only want date precision while js's Date goes up to millis.
So since JS only has the one type, and it's in between FaunaDate and FaunaTime in precision, it seemed better to not use it directly.

test/objects_test.js
andy-faunadb: I think we should require that FaunaTime values are always formatted the same. Otherwise they're not comparable.

eaceaser: well you still need to be able to support passing in an offsetted datetime string into the constructor; it will then parse that and convert it to whatever internal format is appropriate.

andy-faunadb: In python we do `value.replace("+00:00", "Z")` but that's it. Maybe just throw an error if we detect `+` followed by anything other than `Z` or `00:00`? Timezone conversion is a hard problem.

eaceaser: there's got to be an offset parser in javascript out there. its part of the ISO 8601 or whatever it is format. I'd be surprised if it wasn't part of the standard lib.

And its simple in this case, because its a simple offset; it does not contain logical timezone information, such as DST.

andy-faunadb: Once we've parsed, say, a 4 hour offset, we have to add 4 hours to the time, which may mean incrementing the day, which may mean incrementing the month, etc... not something that can be done simply with string manipulation. The Date builtin handles all this math but only with millisecond precision. We could parse out the nanoseconds, then do Date arithmetic, then reintroduce nanos, but would that be worth the complexity? The ability to construct a FaunaTime out of a string is really only meant to be used by us when converting a JSON response.

eaceaser: i guess maybe the issue is more that the date() and time() functions take strings themselves; we should probably have a version that takes a Date object and converts it to the appropriate ISO string? 

the client feels like it needs some form of bridge from the native date/time types to what fauna expects.

andy-faunadb: Why aren't FaunaTime and FaunaDate the bridge we need? They convert to/from Dates, and can be used in and returned from queries.
As for the date and time query functions, I don't know why you'd use them with a string/Date directly, i.e. why using `{"date": "1970-01-02"}` is any better than `{"@date": "1970-01-02"}`.

eaceaser: oh i misread when you said that they were only intended to be used by json conversion; you just meant string construction.

its fine, i don't think this is that big of a deal; if people need to use Date.parse or the equivalent to pass in an offsetted string than thats fine.

andy-faunadb: I changed it to throw if the string doesn't end in 'Z'. This way if people try passing in an offset string it will fail fast.

lib/Client.js
```diff
@@ -304,4 +305,4 @@ function queryStringForLogging(query) {
...
...KVxuICAgIH0pXG4gIH1cbn1cblxuZnVuY3Rpb24gaGFuZGxlUmVzcG9uc2UocmVzcG9uc2UsIHJlc3BvbnNlX29iamVjdCkge1xuICBjb25zdCBjb2RlID0gcmVzcG9uc2Uuc3RhdHVzQ29kZVxuICBpZiAoMjAwIDw9IGNvZGUgJiYgY29kZSA8PSAyOTkpXG4gICAgcmV0dXJuIHJlc3BvbnNlX29iamVjdC5yZXNvdXJjZVxuICBlbHNlXG4gICAgc3dpdGNoIChjb2RlKSB7XG4gICAgICBjYXNlIDQwMDpcbiAgICAgICAgdGhyb3cgbmV3IEJhZFJlcXVlc3QocmVzcG9uc2Vfb2JqZWN0KVxuICAgICAgY2FzZSA0MDE6XG4gICAgICAgIHRocm93IG5ldyBVbmF1dGhvcml6ZWQocmVzcG9uc2Vfb2JqZWN0KVxuICAgICAgY2FzZSA0MDM6XG4gICAgICAgIHRocm93IG5ldyBQZXJtaXNzaW9uRGVuaWVkKHJlc3BvbnNlX29iamVjdClcbiAgICAgIGNhc2UgNDA0OlxuICAgICAgICB0aHJvdyBuZXcgTm90Rm91bmQocmVzcG9uc2Vfb2JqZWN0KVxuICAgICAgY2FzZSA0MDU6XG4gICAgICAgIHRocm93IG5ldyBNZXRob2ROb3RBbGxvd2VkKHJlc3BvbnNlX29iamVjdClcbiAgICAgIGNhc2UgNTAwOlxuICAgICAgICB0aHJvdyBuZXcgSW50ZXJuYWxFcnJvcihyZXNwb25zZV9vYmplY3QpXG4gICAgICBjYXNlIDUwMzpcbiAgICAgICAgdGhyb3cgbmV3IFVuYXZhaWxhYmxlRXJyb3IocmVzcG9uc2Vfb2JqZWN0KVxuICAgICAgZGVmYXVsdDpcbiAgICAgICAgdGhyb3cgbmV3IEZhdW5hSFRUUEVycm9yKHJlc3BvbnNlX29iamVjdClcbiAgICB9XG59XG5cbmZ1bmN0aW9uIHF1ZXJ5U3RyaW5nRm9yTG9nZ2luZyhxdWVyeSkge1xuICByZXR1cm4gcXVlcnkgPyBgPyR7T2JqZWN0LmtleXMocXVlcnkpLm1hcChrZXkgPT4gYCR7a2V5fT0ke3F1ZXJ5W2tleV19YCkuam9pbignJicpfWAgOiAnJ1xufVxuIl19
```
eaceaser: i dont actually know; do we want to include source mapping?

eaceaser:
lgtm

lib/Client.js
andy-faunadb: If people don't want to use the source map then this is just a comment.
It is annoying how changes to code end up changing the diffs like this. I made an issue: #40.

PR #38, opened 2015-11-12
Ignore npm-debug.log

andy-faunadb:


eaceaser:
ship

eaceaser:
ps you can probably commit things like this without review;

PR #41, opened 2015-12-16
add TravisCI integration

sprsquish:


src/model/Model.js
```diff
@@ -380,7 +380,13 @@ export default class Model {
...
 /** Lambda expression for getting an instance Ref out of a match result. */
 function indexRefGetter(index) {
-  return index.values ?
-    arr => query.get(query.select(index.values.length, arr)) :
-    query.get
+  if (index.values) {
+    var idx = -1
+    for (var i = 0; i < index.values.length; i++) {
+      if (index.values[i]['path'] == 'ref') idx = i
+    }
```
sprsquish: If values are defined on the index, refs are no long implicitly covered. They must be covered explicitly. Should we throw an exception here if the index isn't covering refs?

.travis.yml
```diff
@@ -0,0 +1,17 @@
...
+    - FAUNA_DOMAIN=localhost
+    - FAUNA_SCHEME=http
+    - FAUNA_PORT=8443
+jdk:
+  - oraclejdk8
+before_install:
+  - openssl aes-256-cbc -K $encrypted_c0ecaa32cf03_key -iv $encrypted_c0ecaa32cf03_iv -in core.tgz.enc -out core.tgz -d
+  - tar zxf core.tgz
+  - sh install-core.sh
+  - sudo apt-get install -y npm
```
andy-faunadb: Usually installing node installs npm. Are you sure you need this?

sprsquish: Node is pre-installed in the Travis environment. npm is not.

sprsquish: Rather, there's no indication that I can see that it's installed so I'm
ensuring its there.

andy-faunadb:
This should be ready to rebase and merge.

andy-faunadb:
#60 is blocking this.

src/model/Model.js
andy-faunadb: Decided to put solving that problem on hold and removed the methods using this. See #60

PR #42, opened 2015-12-18
Convert to typescript

### LINKED ISSUE #40: Consider removing `dist` from the repo
We had to add this because a new version of npm was updating babel to the wrong version.
Once we're using the new babel we should be able to go back to compiling on install.

andy-faunadb:
Fixes #40


ashfire908:
Seems like `applyDefaults` has a type but the type is never provided when it's called? Does typescript just guess what the type is then?

andy-faunadb:
Yes, typescript will fill in type parameters based on the arguments provided.

andy-faunadb:
Replaced by https://github.com/faunadb/faunadb-js/pull/44

PR #43, opened 2016-01-20
Query improve

andy-faunadb:
Depends on #42


andy-faunadb:
Moved to separate commits.

PR #44, opened 2016-01-27
Babel6

andy-faunadb:


src/model/Model.js
```diff
@@ -66,26 +66,26 @@ export default class Model {
...
           return decoded
         },
-        set(value) {
+        setter(value) {
           const encoded = field.codec.encode(value, this)
           setPath(field.path, encoded, this._current)
         }
       }
-    Object.defineProperty(this.prototype, fieldName, {get, set})
+    Object.defineProperty(this.prototype, fieldName, {get: getter, set: setter})
```
andy-faunadb: Turns out using `{get, set}` as a value is a syntax error.

src/errors.js
```diff
@@ -1,15 +1,5 @@
...
-    super(message)
-    // Subclassing error doesn't work right until node 4.0.
-    // This is a kludge to give it some stack.
-    if (!('stack' in this)) {
-      this.message = message
-      this.stack = new Error(this.message).stack
-    }
-  }
-}
+export class FaunaError extends Error {}
```
andy-faunadb: Since we're not transpiling classes any more we get the proper behavior.

ashfire908:
LGTM

PR #45, opened 2016-01-27
Make queryStringForLogging work for empty hash

andy-faunadb:


andy-faunadb:
Will incorporate this in a commit fixing logging.

PR #48, opened 2016-01-27
Change order of `match` parameters

andy-faunadb:


src/query.js
```diff
@@ -183,7 +183,7 @@ export function delete_expr(ref) {
 }
 
 /** See the [docs](https://faunadb.com/documentation/queries#sets). */
-export function match(terms, index) {
+export function match(index, terms) {
```
ashfire908: In the other clients, terms takes a list of arguments, instead of a array

ashfire908:
LGTM

PR #56, opened 2016-01-28
Add RequestResult class and use it to factor out logging

andy-faunadb:


karenpeng:
@ashfire908 please review this, we need this for the web console project as well~

ashfire908:
LGTM

PR #76, opened 2016-05-25
Convert from ES6+async/await to vanilla ES5

eaceaser:
Note that this is NOT a PR to merge into master. 

This PR does not contain any API or functional changes, it simply converts to vanilla JS and gets tests to pass.

I've moved away all the model / automatic pagination stuff, because it was pretty heavy on the async/await lifestyle. I'll revisit this after we address API changes. 

Todo after this:
* Get docs working
* Get tests across multiple JS runtimes
* API improvements


old/model/_util.js
```diff
@@ -0,0 +1,96 @@
...
+    }
+    return obj;
+  } else {
+    return data;
+  }
+}
+
+/**
+ * Recursively looks for :samp:`path` in :samp:`data`.
+ * e.g. :samp:`get_path(["a", "b"], {"a": {"b": 1}})` should be 1.
```
erickpintor: `get_path` -> `getPath`

eaceaser: ill fix; but stuff in the old/ dirs is probably gonna get deleted eventually :)

old/model/_util.js
```diff
@@ -0,0 +1,96 @@
...
+      return null;
+    }
+    data = data[pathElem];
+  }
+
+  return data;
+}
+
+/**
+ * Opposite of get_path.
```
erickpintor: `get_path` -> `getPath`

old/model/Field.js
```diff
@@ -0,0 +1,13 @@
+var util = require('../_util');
+
+/**
+ * Stores information about a field in a {@link Model}.
+ */
+function Field(opts) {
```
erickpintor: Is this ever used?

test/query_test.js
```diff
@@ -1,452 +1,554 @@
...
+      assert('ref' in instance);
+      assert('ts' in instance);
+      assert.deepEqual(instance.class, classRef);
+    });
+  });
+
+  it('update', function () {
+    return create().then(function (i) {
+      var ref = i.ref;
+      return client.query(query.update(ref, query.quote({ data: { m: 9 } })));
```
erickpintor: Would be nice to call update with a `null` field just to make sure that we can remove a field from an existing instance.

erickpintor:
Is there any reason why just some of the files has `use strict`?

old/model/Field.js
eaceaser: None of the stuff in old/ is being used right now; I wouldn't be surprised if it was internally inconsistent as well

test/query_test.js
eaceaser: Yeah I'm pretty sure we can't right now :) I'll add the test and fix it

eaceaser:
They all probably should at this point

test/query_test.js
eaceaser: I added a test in a later branch to avoid having to rewrite the test multiple times; but I've confirmed that the behavior works as expected

package.json
```diff
@@ -1,102 +1,30 @@
 {
   "name": "faunadb",
-  "version": "0.1.0",
+  "version": "0.2.0",
```
erickpintor: Should we increase the package version now or let we do a separated PR once we decide to publish it?

src/clientLogger.js
```diff
@@ -12,50 +14,67 @@ import {toJSON} from './_json'
...
   function log(str) {
-    out = out + str
+    out = out + str;
   }
 
-  function indent(str) {
-    const indentStr = '  '
-    return str.split('\n').join('\n' + indentStr)
+  log('Fauna ' + method + ' /' + path + _queryString(query) + '\n');
+  log('  Credentials: ' + (auth == null ? 'null' : (auth.user + ':' + auth.pass)) + '\n');
```
erickpintor: Shouldn't we omit the `auth.pass` from logger?

erickpintor:
Question: isn't `Client` too generic? On the other drivers we're prefixing it with `Fauna`.

test/query_test.js
```diff
@@ -1,452 +1,564 @@
...
+var assert = require('chai').assert;
+var errors = require('../src/errors');
+var objects = require('../src/objects');
+var query = require('../src/query');
+var util = require('./util');
+var Promise = require('es6-promise').Promise;
+
+var FaunaDate = objects.FaunaDate,
+  FaunaTime = objects.FaunaTime,
+  Ref = objects.Ref,
```
erickpintor: Ref is quite a common objects that needs to be built. I think all relevant Values should have corespondent functions on the `query` namespace so the user doesn't have to import them all.

package.json
eaceaser: probably increase it now, but append it with '-pre' or something. ill do that

src/clientLogger.js
eaceaser: i remove the auth.pass completely later, but it still emits the secret. I'll change that behavior.

test/query_test.js
eaceaser: yea I end up doing that; when I open a bigger PR we can discuss

eaceaser:
with regards to the client name; I think its OK because node modules are always scoped to their parent unless you explicitly name the object in the script's namespace? So generally people will use 'faunadb.Client'?

PR #79, opened 2016-06-02
Improve API to automatically wrap object literals

eaceaser:
Just like the Ruby client.

### DIFF
--- a/src/_json.js
+++ b/src/_json.js
@@ -1,3 +1,5 @@
+'use strict';
+
 var Expr = require('./Expr');
 var objects = require('./objects');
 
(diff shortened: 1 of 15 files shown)


src/_json.js
```diff
@@ -1,5 +1,6 @@
 'use strict';
 
+var Expr = require('./Expr');
```
erickpintor: Seems like `Expr` is never used on this file.

erickpintor: Forget it... it was removed in another PR.

erickpintor:
🐑  it

PR #80, opened 2016-06-10
Simple pagination helper

eaceaser:
pretty much a JS reimplementation of the pagination helper we have in the ruby client.

Again, note that this does not include any correct documentation; that will happen in a later PR.


src/PageHelper.js
```diff
@@ -0,0 +1,133 @@
...
+    params = {};
+  }
+
+  this.reverse = false;
+  this.params = {};
+  this.cursor = undefined;
+  objectAssign(this.params, params);
+
+  if ('before' in params && 'after' in params) {
+    throw 'Cursor directions "before" and "after"j are exclusive.';
```
erickpintor: Is `"after"j` a typo?

src/PageHelper.js
```diff
@@ -0,0 +1,133 @@
...
+};
+
+PageHelper.prototype._handle_page = function(lambda) {
+  var self = this;
+  return function (page) {
+    lambda(page.data);
+
+    var next_cursor;
+    if (self.reverse) {
+      next_cursor = page.before;
```
erickpintor: Shouldn't `next_cursor` be camel case?

src/PageHelper.js
```diff
@@ -0,0 +1,133 @@
...
+    if (this.reverse) {
+      opts.before = null;
+    }
+  }
+
+  var q = query.paginate(this.set, opts);
+
+  if (this._fauna_functions.length > 0) {
+    this._fauna_functions.forEach(function(lambda) {
+      q = lambda(q);
```
erickpintor: I din't quiet understand this loop. The ideia is that you only keeps the last lamba? Can't we compose them somehow?

src/PageHelper.js
eaceaser: yes, sorry

src/PageHelper.js
eaceaser: No, this will compose them all - since q is being reassigned, it should wrap at each level

erickpintor:
As it stands, `PageHelper` seems to be good for transversing a whole set of pages from a query. I'm just confused if it was meant to help users to move from page to page programatically.

Like: I want to show the first page -> The user click on the "next" button -> I show the second page.

Is that the idea? Does the `PageHelper` api supports it?

eaceaser:
No, but thats difficult to do without manually calling the Paginate function anyways; because most of the time iterative pages will span requests, there won't be any state to maintain.

I see this Helper as assisting with when someone wants to consume the entire set; in reality this won't happen often probably, outside of debugging, so maybe its usefulness is marginal.

Note that this PageHelper is relatively equivalent to the one in the Ruby driver, so anything we learn here we make here should get applied to that driver too.

erickpintor:
In that case. Can we rename the `client.paginate` function at least? I think it is confusing with `query.paginate`. They have the same name but one is for pagination and the other is for transversing...

eaceaser:
yeah. what do you think we should call it?

eaceaser:
maybe just client.pageHelper ..? hrmmmm....

erickpintor:
`client.transverse`? I don't know...

eaceaser:
im starting to wonder whether or not this helper thing is even that useful...

eaceaser:
@ashfire908 do you have any thoughts about the pagination helpers? Do you think that the ruby one provides much value?

ashfire908:
Yes, I think it's valuable. I use it in website + the ruby model to grab pages in a clean way, either one at a time with a given cursor, or iterating over them. It's good for simple usages, though if you need more in-depth use of `paginate`, that should just be used instead (ie the pagination helper does not need to support every use case, just the simple ones).

The way I look at it (at least for the ruby driver) is that if the pagination helper wasn't there, I'd probably just code a quick version of it for each project I need to page over stuff in. I figured a more complete pagination helper directly in the driver would be more useful, hence why I created the one that's in the ruby driver now.

eaceaser:
hmmmm.... I was hoping that the helper would not need to be stateful, but theres no other way to have it be able to iteratively grab subsequent pages

erickpintor:
I don't know. Maybe it's worth having it just to lazily transverse a set of instances. Now that we have a `nextPage`, I wonder how we go back the the previous page....

Perhaps you're right offering a simpler API for transversing and, if the user needs control over pages, he has to go back to the traditional `query.paginate`.

eaceaser:
well i can add a previosuPage too with the same semantics. IDK though. I'm not really 100% into this yet

eaceaser:
ok so I think this is at a decent spot now. LMK what you think

erickpintor:
Looks good to me.

🐑  it!

PR #83, opened 2016-06-24
Capitalize methods

eaceaser:
I think we can figure out a way to combine LambdaExpr / Lambda, but we can do that in another PR


src/query.js
```diff
@@ -128,7 +128,7 @@ function toLambda(value) {
  * @param {module:query~ExprArg} expr
  * @return {Expr}
  * */
-function lambda_expr(var_name, expr) {
+function LambdaExpr(var_name, expr) {
```
erickpintor: Can't we merge `LambdaExpr` and `Lambda` functions?

eaceaser: oh yeah. i forgot to do that. let me do that

erickpintor:
Looks good to me.

🐑  it

PR #87, opened 2016-07-01
Better README and Docs

eaceaser:


README.md
```diff
@@ -1,88 +1,158 @@
...
-We will publish to `npm` soon.
+TODO: Browser source
 
 ### Use
 
-All work with FaunaDB happens through a Client instance.
-All Client methods return [Promises](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise).
-Although it's possible to use Client alone, it's often easier to do work through Model classes.
-For advanced work, the query api (`faunadb/lib/query`) will be useful.
+The [tutorials](http://faunadb.com/documentation/tutorials) in the FaunaDB documentation contain
```
erickpintor: We've recently changed this link to "http://faunadb.com/tutorials" instead.

README.md
```diff
@@ -1,88 +1,158 @@
...
+Note that we are using the `Ref` alias defined above here.
 
-### Setup
+All methods on `faunadb.Client` return [ES6 Promises](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise).
+So, if we wanted to transform the Promise to return the `Ref` of the newly created instance:
 
-    npm install
+```javascript
+createP.then(function(res) {
+  return res.ref;
```
erickpintor: I don't think a `return` is a good example of a promise continuation since the main scope won't have access to the desired ref until the callback function is executed.

I can see people doing things like:

```javascript
var ref  = createP.then(function(res) {
  return res.ref;
});

doSomething(ref); // failed because ref is null / undefined
```

README.md
```diff
@@ -1,88 +1,158 @@
...
 
+Using the helper to page over sets lets the driver handle cursoring and pagination state. For example:
 
-### Test
+```javascript
+var helper = client.paginate(q.Match(Ref('indexes/test_index'), 'example-term'));
+```
 
-    npm run test
+Here, `helper` is an instance of `PageHelper`. The methods `eachPage` and `eachItem`
```
erickpintor: I thought we've removed the `eachItem`...

README.md
```diff
@@ -1,88 +1,158 @@
...
-      "auth": "your_fauna_key"
-    }
+```javascript
+helper.map(function(ref) { return q.Get(ref); }).eachPage(function(page) {
+  console.log(page); // Will now log the retrieved instances.
+});
+```
 
-You can also set the `FAUNA_DOMAIN`, `FAUNA_SCHEME`, `FAUNA_PORT`, and `FAUNA_ROOT_KEY` environment variables.
+[See the JSDocs](https://faunadb.github.com/faunadb-js/PageHelper.html) for more information on the pagination helper.
```
erickpintor: I think it's missing a clear definition on which functions have their effects on the server (`map`, `filter`, e.g) and which ones are local (`eachPage`, `nextPage`, `previousPage`, e.g).

README.md
eaceaser: hmmm, i see your point..... what would you think would be good? another async call?

README.md
eaceaser: I added this to the PageHelper docs

erickpintor:
Looks good to me. 🐑  it.

PR #88, opened 2016-07-13
Final JS Client Code Review

eaceaser:
make sure nothings busted. I'll fix the NPM paths and stuff in the README in the meantime.


README.md
```diff
@@ -1,97 +1,177 @@
...
 
-`domain`, `scheme`, and `port` are optional and will default to FaunaDB cloud.
-So to test with cloud, `testConfig.json` is as simple as:
+For example, to retrieve the matched instances:
 
-    {
-      "auth": {"user": "me@example.com", "pass": "swordfish"}
-    }
+```javascript
+helper.map(function(ref) { return q.Get(ref); }).eachPage(function(page) {
```
freels: eachPage is inconsistent with each above

README.md
```diff
@@ -1,97 +1,177 @@
...
+more information on the pagination helper.
 
-Note that when testing in cloud, builtin tests for database, key, and custom field will fail as cloud accounts do not have access to that functionality.
+## Client Development
 
-Testing uses a require hook, so building isn't necessary.
+Run `npm install` to install dependencies.
 
-For fuller testing, run `./test-all.sh`, which uses [nvm](https://github.com/creationix/nvm).
+### Code
```
freels: Perhaps better labeled Dependencies

freels: nm, i guess it's ok.

src/Expr.js
```diff
@@ -0,0 +1,63 @@
...
+Expr.wrap = function(obj) {
+  if (obj === null) {
+    return null;
+  } if (obj instanceof Expr) {
+    return obj;
+  } else if (obj instanceof Array) {
+    return new Expr(obj.map(function (elem) {
+      return Expr.wrap(elem);
+    }));
+  } else if (typeof obj === 'object') {
```
freels: i hate js

src/PageHelper.js
```diff
@@ -0,0 +1,234 @@
...
+  }
+
+  return this.client.query(q);
+};
+
+/**
+ * @private
+ * @returns {PageHelper}
+ */
+PageHelper.prototype.clone = function() {
```
freels: if this is private should it be named `_clone`?

src/Expr.js
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * Wraps an object as an Expression. This will automatically wrap any bare objects with
+ * the appropriate {@link object} escaping.
+ * @param {Object} obj
+ *  The object to be wrapped as an Expression.
+ * @returns {Expr}
+ *   The expression wrapping the provided object.
+ */
+Expr.wrap = function(obj) {
```
freels: in ruby this wraps lambdas as well. is that possible here, so that raw functions will be converted to Lambdas?

src/query.js
```diff
@@ -1,335 +1,708 @@
...
 
-/** See the [docs](https://faunadb.com/documentation/queries#misc_functions). */
-export function equals(...values) {
-  return {equals: varargs(values)}
+/**
+ * See the [docs](https://faunadb.com/documentation/queries#misc_functions).
+ *
+ * @return {Expr}
+ */
+function NextId() {
```
freels: `NextID`? My intuition is to prefer consistency across clients if applicable, but it's not a big deal.

src/query.js
```diff
@@ -1,335 +1,708 @@
...
+ * @return {Expr}
+ */
+function Not(boolean) {
+  return new Expr({ not: Expr.wrap(boolean) });
 }
 
 // Helpers
 
 /** Adds optional parameters to the query. */
 function params(mainParams, optionalParams) {
```
freels: is this function exported? should be hidden if so.

src/query.js
```diff
@@ -338,5 +711,77 @@ function params(mainParams, optionalParams) {
  * `query.add([1, 2])` will work as well as `query.add(1, 2)`.
  */
 function varargs(values) {
```
freels: ditto

src/query.js
```diff
@@ -338,5 +711,77 @@ function params(mainParams, optionalParams) {
...
+
+function defaults(param, def) {
+  if (param === undefined) {
+    return def;
+  } else {
+    return param;
+  }
+}
+
+module.exports = {
```
freels: i see i've answered my own question. :)

freels:
awesome.

src/Expr.js
eaceaser: its already being done in the query#Lambda() function, but we can move that logic here too probably

src/query.js
eaceaser: what do you mean hidden? from the jsdocs? the jsdocs by default don't expose functions that aren't exported

src/Expr.js
freels: yeah, calling Lambda from here is fine. I mean automatically invoking it here so you can pass raw functions in lieu of calling Lambda.

freels: and removing the special handling of lambda args in the query expression constructors.

eaceaser: word

eaceaser: ohhh i remember why I had to do this this way. because of the circular dependency between Expr and query, I either have to export Expr from query (which makes all the exporting and typing more complex cuz javascript, as everything now is query#Expr as opposed to a bare Expr class), or re-implement the Var function in Expr.

I guess I can do the latter... but I don't really like that either. I'll see if I can come up with a better way.

freels: I guess the logic could live in here and Lambda just call Expr.wrap?

eaceaser: the problem is the lambda code uses the Var() function, so I'd have to reimplement that. Its not a problem (its oneline), but it does technically reduplicate that logic.

src/Client.js
```diff
@@ -1,172 +1,240 @@
...
-  }
+/**
+ * Executes a query via the FaunaDB Query API.
+ * See the [docs](https://faunadb.com/documentation/queries),
+ * and the query functions in this documentation.
+ * @param expression {Expr}
+ *   The query to execute. Created from query functions such as {@link add}.
+ * @return {external:Promise<Object>} FaunaDB response object.
+ */
+Client.prototype.query = function (expression) {
```
freels: This should pass `expression` through Expr.wrap.

eaceaser: wait, why? Then you couldn't pass in a raw query obj that you've crafted yourself, because it'd get escaped with { object }

erickpintor: I think that is the idea. It keeps the echo principle. 
The rest of the drivers doesn't allow people to create their own expressions neither.

freels:
shipit!

PR #91, opened 2016-07-14
Add package links

eaceaser:
and cleanup package.json

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,17 +1,16 @@
 {
   "name": "faunadb",
   "version": "0.2.0",
-  "description": "FaunaDB (faunadb.com) client for node",
+  "description": "FaunaDB Javascript driver for Node.JS and Browsers",
   "homepage": "https://faunadb.com",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,17 +1,16 @@
...
   "repository": "faunadb/faunadb-js",
   "license": "MPL-2.0",
+  "keywords": [ "database", "fauna", "official", "driver" ],
+  "bugs": { "url": "https://github.com/faunadb/faunadb-js/issues" },
   "files": [
     "index.d.ts",
     "src/"
   ],
-  "engines": {
-    "node": ">= 4.0.0"
```
erickpintor: Isn't this the minimal supported version? What happens if we remove it?

eaceaser: it works in 0.12.x as well. i figured taking it out was actually more useful than leaving it in, considering that i dont think anything really honors that

erickpintor:
🐑  it.

PR #95, opened 2016-09-29
Update doc links

ashfire908:
Part of https://github.com/faunadb/sales-engineering/issues/348.

Does not make any code changes.

### DIFF
--- a/src/Client.js
+++ b/src/Client.js
@@ -23,7 +23,7 @@ var Promise = require('es6-promise').Promise;
  *
  * Users will mainly call the {@link Client#query} method to execute queries.
  *
- * See the [FaunaDB Documentation](https://faunadb.com/documentation) for detailed examples.
+ * See the [FaunaDB Documentation](https://fauna.com/documentation) for detailed examples.
  *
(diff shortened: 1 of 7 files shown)


src/Client.js
```diff
@@ -42,7 +42,7 @@ var Promise = require('es6-promise').Promise;
  *   HTTP scheme to use.
  * @param {?number} options.port
  *   Port of the FaunaDB server.
- * @param {?string} options.secret FaunaDB secret (see [Reference Documentation](https://faunadb.com/documentation/objects#keys))
+ * @param {?string} options.secret FaunaDB secret (see [Reference Documentation](https://fauna.com/documentation/objects#keys))
```
erickpintor: I can't remember if we're also renaming `FaunaDB` to `Fauna`. Did you rename those on other drivers?

erickpintor: I've checked other PRs. I don't think we are. That's fine, I guess.

ashfire908: The database is still called FaunaDB. I'm only updating links to the website. Did I do more than that somewhere?

erickpintor: No. I was the one that was confused. You're fine.

PR #101, opened 2016-11-02
Restrict functions arity

erickpintor:
Closes: https://github.com/faunadb/sales-engineering/issues/379

### DIFF
--- a/src/query.js
+++ b/src/query.js
@@ -32,6 +32,7 @@ var objectAssign = require('object-assign');
  * @return {Expr}
  */
 function Ref() {
+  arity.min(1, arguments);
   var args = argsToArray(arguments);
   return new (values.Ref.bind.apply(values.Ref, [null].concat(args)));
 }
@@ -46,6 +47,7 @@ function Ref() {
  * @return {Expr}
  * */
 function Let(vars, in_expr) {
+  arity.exact(2, arguments);
   return new Expr({ let: wrapValues(vars), in: wrap(in_expr) });
 }
 
@@ -56,6 +58,7 @@ function Let(vars, in_expr) {
  * @return {Expr}
  * */
 function Var(varName) {
+  arity.exact(1, arguments);
   return new Expr({ var: wrap(varName) });
 }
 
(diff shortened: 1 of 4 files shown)


src/query.js
```diff
@@ -744,11 +801,27 @@ function Or() {
...
 
 // Helpers
 
+/**
+ * @ignore
+ */
+function arity(min, max, message, args) {
+  if ((min !== null && args.length < min) || (max !== null && args.length > max)) {
+    throw new errors.InvalidValue(
+      'Function require ' + message + ' arguments but ' + args.length + ' were given.');
```
eaceaser: typo: require => requires

src/query.js
```diff
@@ -744,11 +801,27 @@ function Or() {
...
 }
 
 // Helpers
 
+/**
+ * @ignore
+ */
+function arity(min, max, message, args) {
+  if ((min !== null && args.length < min) || (max !== null && args.length > max)) {
+    throw new errors.InvalidValue(
```
eaceaser: On second thought, should we just create another exception class and expose the arity information there, so we don't have to do regex string matching on the message to determine if the correct error was thrown?

jchris: I implemented that in dafd1e8 but I'm not attached to it either way.

erickpintor:
@ed can you do another pass?

erickpintor:
Is there anything else missing for this PR?

jchris:
The only thing I can think of is that the tests don't actually use the InvalidArity features, and the message composition for that error type is still implemented in query.js as a helper, instead of being extracted to the error class. It seemed there was some additional refactoring needed here that would be obvious to the first person who starts to modify the InvalidArity type, so maybe it's better to wait and do that later? I'm happy to write the code either way I guess it depends on the prevailing winds as far as late-binding in our client code.

erickpintor:
I have moved the error message composition to the error class.

Regarding make usage of arity features on the tests, I'm a little hesitant to make arity function public. It doesn't seems to be useful as a public interface at the driver since they are safe guards due to the lack of a more strict language option.

eaceaser:
lgtm

PR #104, opened 2016-12-07
update URL

jchris:
I heard the endpoint was supposed to be `db.fauna.com`


erickpintor:
Are both `could` and `db` pointing to the same cluster?

sprsquish:
Yes, `cloud` and `db` are aliases

erickpintor:
I would just add an entry on CHANGE.log. Besides that, it looks good.

PR #107, opened 2017-01-23
Add support for the true `ref` function

eaceaser:
I'm not sure if we should continue to have the driver internally join ref parts by `/` anymore anywhere, actually. It should probably just have the standard Ref() constructor, and the `{ ref: id: }` function form.

### DIFF
--- a/test/query_test.js
+++ b/test/query_test.js
@@ -520,8 +520,8 @@ describe('query', function () {
   });
 
   it('ref', function() {
-    var id = refN1.id.toString();
-    return assertQuery(query.Ref(classRef, query.Concat([id.charAt(0), id.substr(1)])), refN1);
+    var ref = query.Ref(classRef.value + "/123456");
+    return assertQuery(query.Ref(classRef, query.Concat(["123", "456"])), ref);
   });
 
   // Check arity of all query functions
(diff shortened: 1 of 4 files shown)


test/query_test.js
```diff
@@ -519,12 +519,17 @@ describe('query', function () {
     return Promise.all([p1, p2]);
   });
 
+  it('ref', function() {
+    var id = refN1.id.toString();
+    return assertQuery(query.Ref(classRef, query.Concat([id.charAt(0), id.substr(1)])), refN1);
```
erickpintor: I didn't understand this assertion. Can't we just use literals to make it more clear?

test/util.js
```diff
@@ -106,7 +106,7 @@ function unwrapExprValues(obj) {
 
 var rootClient = getClient({ secret: testConfig.auth });
 var dbName = 'faunadb-js-test-' + randomString();
-var dbRef = Ref('databases', dbName);
+var dbRef = Ref(['databases', dbName].join('/'));
```
erickpintor: What about just `"databases/" + dbName`?

test/query_test.js
eaceaser: for some reason I thought if the ref was invalid it'd error out, but I guess that's not true

test/util.js
eaceaser: good question

PR #109, opened 2017-02-09
adds support for key_from_secret function

marrony:
close https://github.com/fauna/sales-engineering/issues/493

### DIFF
--- a/src/query.js
+++ b/src/query.js
@@ -274,6 +274,17 @@ function Get(ref, ts) {
 
 /**
  * See the [docs](https://fauna.com/documentation/queries#read_functions).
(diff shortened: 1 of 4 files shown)


src/query.js
```diff
@@ -274,6 +274,17 @@ function Get(ref, ts) {
 
 /**
  * See the [docs](https://fauna.com/documentation/queries#read_functions).
+ *
+ * @param {?module:query~ExprArg} secret
```
erickpintor: This is not an optional parameters, is it?

marrony: damn, copy-and-past

PR #111, opened 2017-02-13
Adds support for @bytes type

marrony:
depends on https://github.com/fauna/faunadb-js/pull/110


src/values.js
```diff
@@ -216,10 +217,41 @@ FaunaDate.prototype.toJSON = function()  {
...
+ *    If string it must be base64 encoded and it's converted to Uint8Array
+ * @extends module:values~Value
+ * @constructor
+ */
+function Bytes(value) {
+  if (value instanceof ArrayBuffer) {
+    this.value = new Uint8Array(value);
+  } else if (typeof value === 'string') {
+    this.value = base64.toByteArray(value);
+  } else {
```
erickpintor: Since this is a public interface, it would be safer to if we throw an exception when people try to create a `Bytes` instance without a valid binary type.

PR #117, opened 2017-03-06
force Content-Type to json and serialize data before send

marrony:
This PR forces the `Content-Type` of request to `application/json` and serialize data to string before passing it to superagent so the library don't have to infer the data type.

This fixes the case when we just send primitive types and expect faunadb to echo these values.

### DIFF
--- a/src/Client.js
+++ b/src/Client.js
@@ -200,9 +200,8 @@ Client.prototype._performRequest = function (action, path, data, query) {
     rq.query(query);
   }
 
-  if (data) {
-    rq.send(data);
-  }
+  rq.type('json');
+  rq.send(JSON.stringify(data));
 
   if (this._secret) {
     rq.set('Authorization', secretHeader(this._secret));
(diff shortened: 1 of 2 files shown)


src/Client.js
```diff
@@ -200,9 +200,8 @@ Client.prototype._performRequest = function (action, path, data, query) {
     rq.query(query);
   }
 
-  if (data) {
-    rq.send(data);
-  }
+  rq.type('json');
+  rq.send(JSON.stringify(data));
```
erickpintor: I think we need to test for `undefined`. The previous if statement prevented `undefined` to be sent to the server.

marrony: IDK, send `undefined` and send no data the outcome would be the same

```json
'{"errors":[{"position":[],"code":"invalid expression","description":"No query provided."}]}'
```

erickpintor: Ok. I thought it wold break with an exception or something.

PR #120, opened 2017-03-20
Accept lambdas for let expression

erickpintor:
This is a syntax sugar proposal. It avoid `q.Var` expressions when using `q.Let`.

Before:
```javascript
q.Let({ x: 1, y: 2 }, q.Add(q.Var("x"), q.Var("y")))
```

Now:
```javascript
q.Let({ x: 1, y: 2 }, (x, y) => q.Add(x, y))
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
-## 1.0.2-SNAPSHOT
+## 1.1.0-SNAPSHOT
+- Accept lambdas at Let function
 
 ## 1.0.1 (March 21, 2017)
 - Fix wrap of Select's default parameter in order to allow other types rather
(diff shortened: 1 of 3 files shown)


marrony:
How it will behave with minification? I see you are using `annotate(in_expr)` to get the name of the arguments.

erickpintor:
Good call @marrony. I've changed the code.

marrony:
just being nitpicking, how `Object.keys` is ordered? Declaration order? What if it return in a different order, it will pass the `Var` objects to lambda in a not expected order.

erickpintor:
It has the same order as the `for..in` statement, which is implementation specific. Although, as far as I can tell, `for..in` in literal objects return the keys in the same order as they were defined.

PR #125, opened 2017-03-22
Use node scope for ping test

erickpintor:
`all` is too slow for the purpose of this test.


erickpintor:
Yes. `all` is usually slow for this kind of things. The idea is to test that we *can* ping, and not that all cloud is up.

Issue #130, opened 2017-06-30
Default class for types is unnecessary


martimatix:
I was playing around with TypeScript and the faundb js client today and no matter what I tried, the following line was giving me problems in my text editor.

```typescript
import * as faunadb from 'faunadb';
const client = new faunadb.Client({
  secret: credentials.faunadbSecret
});
```
The error message was that the property 'Client' does not exist on faunadb.

Once I went into [the type definitions](https://github.com/fauna/faunadb-js/blob/master/src/types/Client.d.ts) for the client and changed the definition from `export default class Client` to `export class Client`, things started working. 

Can you confirm that the above change should be made to this repo (as well as all the other *default* class types) or am I missing something in my implementation?

erickpintor:
Hi @martimatix. Thanks for reporting this. You're absolutely right. The types are not correctly exported.

This will be fixed as soon as we release: https://github.com/fauna/faunadb-js/pull/131

erickpintor:
Fixed at version `1.1.2`. It's already released to npm.

PR #131, opened 2017-07-03
Fix typescript declaration file to include default types

erickpintor:
Fixes: #130 

Context: Typescript `export * from "./file"` exports everything except for the `default` export. Therefore, `Client` is not available to the typescript compiler, although present at runtime.


eaceaser:
lgtm

PR #135, opened 2017-08-25
DRIV-2: Adds support for @query types (user defined functions)

marrony:


src/query.js
```diff
@@ -190,6 +190,46 @@ function _lambdaExpr(var_name, expr) {
...
+ * Query(function (a, b) { return Add(a, b) })
+ * ```
+ *
+ * @param {module:query~ExprArg|function} lambda
+ * @return {Expr}
+ * */
+function Query(lambda) {
+  arity.exact(1, arguments);
+  if (lambda instanceof Function)
+    lambda = _lambdaFunc(lambda);
```
erickpintor: Why do you need to call `_lambdaFunc` specifically? Seems like `wrap` already does that.

src/types/query.d.ts
```diff
@@ -14,6 +14,8 @@ export module query {
   export function Object(fields: ExprArg): Expr;
   export function Lambda(f: Lambda): Expr;
   export function Lambda(var_name: ExprArg, expr: ExprArg): Expr;
+  export function Call(ref: ExprArg, ...args: ExprArg[]): Expr;
+  export function Query(lambda: ExprArg): Expr;
```
erickpintor: Isn't `lambda` supposed to be of type `Lambda`?

marrony: not necessarily, I could be a return of `Lambda()` function as well, so I will add a overload with `Lambda` type

src/types/query.d.ts
```diff
@@ -14,6 +14,9 @@ export module query {
   export function Object(fields: ExprArg): Expr;
   export function Lambda(f: Lambda): Expr;
   export function Lambda(var_name: ExprArg, expr: ExprArg): Expr;
+  export function Call(ref: ExprArg, ...args: ExprArg[]): Expr;
+  export function Query(lambda: Lambda): Expr;
+  export function Query(lambda: ExprArg): Expr;
```
erickpintor: I don't think you need this overload considering the signature of `Map`, `Foreach`, and `Filter`. Since it's JS, I think the intent was to enforce the usage of native lambdas.

marrony: this is a typescript thing, I think the compiler needs this overloaded version to type check the callers.

erickpintor: If so, don't we need the same for the other functions that accept lambdas? Also, can't we use the `ExprArg|Lambda` syntax here?

erickpintor: I'm asking because this function has a different signature than its similar siblings.

marrony: well, removing this, ts users won't be able to to things like `Query(Lambda("x", Var("x")))` only `Query(x => x)`

marrony: not that this is a problem :D

erickpintor: I prefer the latter, looks cleaner. That's what I meant by enforcing native lambdas. Regardless of your choice, let's make sure every function that accepts lambda works the same.

marrony: Ok, I removed that overloaded version

PR #136, opened 2017-09-19
DRIV-11: Adds support for recursive references

marrony:


src/Client.js
```diff
@@ -95,72 +95,12 @@ Client.prototype.paginate = function(expression, params) {
 };
 
 /**
- * Issues a HTTP `GET` request via the legacy REST API.
```
erickpintor: Removing this means that we should release a major version and also document it as part of the CHANGELOG.

src/_json.js
```diff
@@ -20,7 +20,15 @@ function json_parse(_, val) {
   if (typeof val !== 'object' || val === null) {
     return val;
   } else if ('@ref' in val) {
-    return new values.Ref(val['@ref']);
+    var ref = val['@ref'];
+
+    if (!('class' in ref) && !('database' in ref))
+      return values.Native.fromName(ref['id']);
+
+    var cls = json_parse(_, ref['class']);
```
erickpintor: I think it's preferable to call `parseJSON`, instead. I don't feel comfortable passing the `_` variable further.

src/values.js
```diff
@@ -34,62 +35,54 @@ util.inherits(Value, Expr);
...
-  /**
-   * The string value of the ref.
-   *
-   * @type {string}
-   */
-  this.value = parts.join('/');
+function Ref(id, clazz, database) {
+  this.value = {};
+
+  if (id) this.value['id'] = id;
```
erickpintor: Shouldn't we throw an exception if no `id`?

src/values.js
```diff
@@ -99,7 +92,9 @@ Ref.prototype.toJSON = function() {
 
 /** @ignore */
 Ref.prototype.toString = function() {
-  return this.value;
+  var cls = this.class !== undefined ? ', class=' + this.class.toString() : '';
+  var db = this.database !== undefined ? ' database = ' +  this.database.toString() : '';
```
erickpintor: class and database format should be equal.

src/query.js
```diff
@@ -1053,6 +1173,14 @@ module.exports = {
   Database: Database,
   Index: Index,
   Class: Class,
+  Function: FunctionFn,
```
erickpintor: I almost forgot. We need to add this functions to `query.d.ts` too.

src/_json.js
marrony: `parseJSON` receives a string not an object

src/values.js
```diff
@@ -99,7 +92,9 @@ Ref.prototype.toJSON = function() {
 
 /** @ignore */
 Ref.prototype.toString = function() {
-  return this.value;
+  var cls = this.class !== undefined ? ', class=' + this.class.toString() : '';
+  var db = this.database !== undefined ? ' database=' +  this.database.toString() : '';
+  return 'Ref(id=' + this.id + cls + db + ')';
```
erickpintor: I just realized this strings are a big part of our dashboard repl. People can right now just copy and past references' representations back into the repr and run them. I think we should keep this behaviour by making the string representation of a Ref type a valid Ref constructor.

marrony: this will be tricky, wouldn't be easier to the dashboard manage this?

erickpintor: That's a good point. I'll open a ticket for the dashboard work.

PR #140, opened 2017-11-22
Add typescript to dev deps, fix typedefs

eaceaser:
There may be some further improvements we can make to our typedefs. But this gives us the ability to run tsc on our typedefs as part of our build, and, also makes our typedefs work with modern typescript.

### DIFF
--- a/src/types/Client.d.ts
+++ b/src/types/Client.d.ts
@@ -14,14 +14,14 @@ export interface ClientConfig {
 export default class Client {
   constructor(opts?: ClientConfig);
 
-  query(expr: Expr): Promise<Object>;
+  query(expr: Expr): Promise<object>;
   paginate(expr: Expr): PageHelper;
 
-  get(path: string, query?: Object): Promise<Object>;
-  post(path: string, data: Object): Promise<Object>;
-  put(path: string, data: Object): Promise<Object>;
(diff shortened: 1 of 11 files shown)


src/types/Client.d.ts
```diff
@@ -14,14 +14,14 @@ export interface ClientConfig {
...
-  query(expr: Expr): Promise<Object>;
+  query(expr: Expr): Promise<object>;
   paginate(expr: Expr): PageHelper;
 
-  get(path: string, query?: Object): Promise<Object>;
-  post(path: string, data: Object): Promise<Object>;
-  put(path: string, data: Object): Promise<Object>;
-  patch(path: string, data: Object): Promise<Object>;
-  delete(path: string): Promise<Object>;
+  get(path: string, query?: object): Promise<object>;
```
marrony: those functions doesn't exist anymore :( we should remove this from here

marrony: there are only `query`, `ping` and `paginate`

eaceaser:
@marrony removed. Is there anything in the `query` module's typedefs that needs changing that you know about?

marrony:
nothing that I'm aware of, I tried to change typescript according to js every change I made, I just forgot to change the client interface.

eaceaser:
@marrony is it good to merge now?

PR #155, opened 2018-05-21
properly check for ref equality

sprsquish:


src/values.js
```diff
@@ -115,7 +115,10 @@ Ref.prototype.inspect = function() {
  * @returns {boolean}
  */
 Ref.prototype.equals = function(other) {
-  return other instanceof Ref && this.value === other.value;
+  return (other instanceof Ref) &&
+    this.id == other.id &&
```
erickpintor: Why not `===` in this case?

src/values.js
```diff
@@ -115,7 +115,10 @@ Ref.prototype.inspect = function() {
...
+    (this.class === undefined || this.class.equals(other.class)) &&
```
erickpintor: What if `other` nas a `class` or `database` field and `this` doesn't. Wouldn't that make them different?
The `id` field is probably "equal enough" but, technically, it's best to assert the entire data structure, right?

sprsquish: The `equals` check should fail if `other` doesn't have a class or database.

We were asserting the entire structure before, but I don't think it was using the defined `equals` method. As such, Refs with the same structure were failing the check.

erickpintor: It think what happened is that the `value` field was the only internal representation of `Ref` when the `equals` method was created. At the time, the `value` field was a string containing the reference id. 

While adding support for recursive references, the `value`'s field type was changed but, the implementation of `equals` wasn't.

My previous comment was more about the fact that on the line 120, if `this.class` is `undefined`, the condition is true regardless if the `other.class` is not `undefined`, which in that case would mean that both references are not equal.

The same goes for database check at line 121.

Does that make sense?

sprsquish: Oh, derp. You're right. I see what you're saying now.

src/values.js
```diff
@@ -115,7 +115,14 @@ Ref.prototype.inspect = function() {
  * @returns {boolean}
  */
 Ref.prototype.equals = function(other) {
-  return other instanceof Ref && this.value === other.value;
+  return (other instanceof Ref) &&
+    this.id === other.id &&
+    (this.class instanceof Ref) &&
+    (other.class instanceof Ref) &&
+    this.class.equals(other.class) &&
+    (this.database instanceof Ref) &&
```
erickpintor: I think this will still fail. I was told `class` and `database` are not required fields.

```
$ curl -u secret: http://localhost:8443/ -H 'X-FaunaDB-API-Version: 2.1' -d '{ "database": "db" }'
{
  "resource": {
    "@ref": { "id": "db", "class": { "@ref": { "id": "databases" } } }
  }
}
```

sprsquish: gah. sorry. I'm jumping around too much.

PR #156, opened 2018-05-24
Set api version to 2.0

marrony:
this creates the branch `release-1.x` right before we introduce recursive references and set to the version 2.1.


erickpintor:
Changes in the change log should also go to master.

PR #164, opened 2018-07-20
Pretty print Fauna objects

marrony:
Here are examples of the output generated by this change:
```javascript
Database("sub-db", Database("db"))
Index("spells_by_element_with_name")
Class("spells")
Class("cls", Database("db"))
SetRef({ match: Index("all_spells") })
Ref(Class("spells"), "201377948995944972")
Bytes("1234")
Date("2018-07-20T22:12:51.443072Z")
Date("2018-01-01")

{ ref: Function("fn", Database("db")),
  ts: 1532123046028578,
  name: 'fn',
  body: Query({ lambda: 'x', expr: { var: 'x' } }) }

Ref(Keys(), "159541672656502790")
```

### DIFF
--- a/src/values.js
+++ b/src/values.js
@@ -304,7 +304,7 @@ function Query(value) {
 util.inherits(Query, Value);
 
 wrapToString(Query, function() {
-  return 'Query(' + stringify(this.value) + ')';
+  return 'Query(' + Expr.toString(this.value) + ')';
 });
 
 /** @ignore */
--- a/src/values.js
+++ b/src/values.js
@@ -5,6 +5,9 @@ var errors = require('./errors');
 var Expr = require('./Expr');
 var util = require('util');
 
+var customInspect = util && util.inspect && util.inspect.custom;
+var stringify = (util && util.inspect) || JSON.stringify;
+
 /**
  * FaunaDB value types. Generally, these classes do not need to be instantiated
  * directly; they can be constructed through helper methods in {@link module:query}.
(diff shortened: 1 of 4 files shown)


alvarofauna:
Two questions: 

1. Why `Query` becomes `QueryV`.
2. @freels says we want to avoid seeing things like `[object Object]` or `[Array array]`.  Would that be covered with these changes?

marrony:
that won't change the behavior of native `Object` when it's very deep in the hierarchy like in this case
```javascript
{ a: [ 1, 2, 3, 4, 5, 6, 7, 8, 9, 0, { nested: [Object] } ],
  b: { c: 'str' },
  d: Ref(Class("cls"), "123") }
```
it changes only the fauna types, idk if we can add a custom inspector for native objects, will have to search online

marrony:
the example above was produced by the query
```javascript
client.query({a: [1, 2, 3, 4, 5, 6, 7, 8, 9, 0, {nested: {key: "value"}}], b: {c: "str"}, d: q.Ref(q.Class("cls"), "123")}).then(function(res) {
    console.log(res);
})
```

alvarofauna:
It should be doable with the JSON result itself: 

https://github.com/fauna/dashboard/blob/master/packages/dashboard-base/src/dataset/stringify.js#L6

I used that for the fauna-shell

marrony:
Ok, I did some research and came to a conclusion that this is not driver's responsibility. The driver returns a data to the user and it's up to the user print it. Calling `JSON.stringify` will not do it right because it will ignore the custom inspect function.

The fauna shell might call:
```javascript
var util = require('util');

console.log(util.inspect(result, {depth: null}));
```
to pretty-print the `result` without any depth limit. There are many options you can use to control the formatting like line breaks and etc. https://nodejs.org/api/util.html#util_util_inspect_object_options

alvarofauna:
Yeah I agree, I if the user doesn't want to see `[object Object]` or `[Array array]` they should take care of that.

alvarofauna:
It's fine for me BTW to merge.

freels:
We need to make sure that the full result is recursively pretty-printed out in the doc example responses then.

src/values.js
```diff
@@ -90,23 +93,43 @@ Ref.prototype.toJSON = function() {
...
+  classes: "Class",
+  databases: "Database",
+  indexes: "Index",
+  functions: "Function"
 };
 
-/** @ignore */
-Ref.prototype.valueOf = function() {
-  return this.value;
+Ref.prototype._toString = function() {
```
erickpintor: Why not define the `toString` instead?

src/values.js
```diff
@@ -160,10 +183,16 @@ function SetRef(value) {
...
 /** @ignore */
-SetRef.prototype.inspect = function() {
-  return 'SetRef(' + JSON.stringify(this.value) + ')';
+SetRef.prototype.toString = function() {
+  return 'SetRef(' + stringify(this.value) + ')';
 };
 
+SetRef.prototype.inspect = SetRef.prototype.toString;
+
+if (customInspect) {
```
erickpintor: Maybe you could wrap everything in a function like `wrapToString(Ref, function() { ... }` just so we don't forget to always add `toString`, `inspect, and `customInspect`?

marrony:
@erickpintor I did what you suggested

test/values_test.js
```diff
@@ -120,4 +121,37 @@ describe('Values', function() {
...
+    assertPrint(new Ref('idx', values.Native.INDEXES, db), 'Index("idx", Database("db"))');
+    assertPrint(new Ref('fn', values.Native.FUNCTIONS, db), 'Function("fn", Database("db"))');
+    assertPrint(new Ref('key', values.Native.KEYS, db), 'Ref(Keys(Database("db")), "key")');
+
+    assertPrint(new FaunaTime('1970-01-01T00:00:00.123456789Z'), 'Date("1970-01-01T00:00:00.123456789Z")');
+    assertPrint(new FaunaDate('1970-01-01'), 'Date("1970-01-01")');
+
+    assertPrint(new SetRef({'match': new Ref('idx', values.Native.INDEXES)}), 'SetRef({ match: Index("idx") })');
+    assertPrint(new Bytes('1234'), 'Bytes("1234")');
+    assertPrint(new Query({'lambda': 'x', 'expr': {'var': 'x'}}), 'Query({ lambda: \'x\', expr: { var: \'x\' } })');
```
erickpintor: Is it worth making this look like a function?

marrony: What do you mean? the lambda inside `Query(...)`? So, since @freels said he didn't want to see `[object Object]` in our doc responses, I just stringify the internal value.

erickpintor: I though it'd be nice to see lambda as a JS function. Like `Query(x => x)`, if possible.

freels: If not that, then `Lambda` constructor syntax.

alvarofauna: The `Lambda` syntax could later be copy pasted into a shell. So it can help people reading the examples

marrony: Well, that was not the best code ever, but do the job

PR #165, opened 2018-07-25
Make string representation copy-and-pastable for fauna-shell

marrony:


src/query.js
```diff
@@ -86,7 +96,7 @@ function At(timestamp, expr) {
 function Let(vars, in_expr) {
   arity.exact(2, arguments);
 
-  if (in_expr instanceof Function) {
+  if (typeof in_expr === 'function') {
```
erickpintor: I think someone hit a wried JS runtime! :D

PR #167, opened 2018-07-27
Add support for X-Last-Seen-Txn.

bitbckt:
I have no clue what I'm doing here.

### DIFF
--- a/src/Client.js
+++ b/src/Client.js
@@ -65,6 +65,7 @@ function Client(options) {
   this._timeout = Math.floor(opts.timeout * 1000);
   this._secret = opts.secret;
   this._observer = opts.observer;
+  this._lastSeen = null;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


src/Client.js
```diff
@@ -125,6 +126,16 @@ Client.prototype._execute = function (action, path, data, query) {
       response.text, responseObject, response.status, response.header,
       startTime, endTime);
 
+    if ('x-last-seen-txn' in response.header) {
+        var time = parseInt(response.header['x-last-seen-txn'], 10);
```
marrony: that could be an issue on javascript as it cannot represent a java long value 😞

erickpintor: Yep.

bitbckt: I hate this language.

erickpintor: Idk... split the string in half and compare both sides? _(runs aways really quickly)_

We might need to pick a library for this....

marrony: Seems like `BigInt` will be part of js standard some day in the future, nodejs and chrome already support it, but safari don't

bitbckt: I think this is safe... 2^53 micros since the epoch is 11:47:34 pm UTC Tuesday, June 5, 2255. That seems sufficient...

erickpintor: Who would say math was gonna give us a simpler solution, hum?!

PR #168, opened 2018-08-01
Release 2.5.0

bitbckt:


erickpintor:
The PR looks good. Although I'd feel more comfortable if we fix the failing tests for node 4 and 6. If we still support them. @marrony what do you think? Those missing polyfills might happen in the browser too, right?

marrony:
🤦‍♂️ I don't believe node 4 and 6 doesn't have support for `Object.values`. I was so obvious to me given it support `Object.keys`

erickpintor:
I think there are a few drivers that also doesn't support it. I think we should fix it.

bitbckt:
Is there a ticket for this problem?

marrony:
@erickpintor What other drivers?

erickpintor:
I'm sorry. I meant a few *browsers*.

sprsquish:
pulled internally

PR #174, opened 2019-01-22
make client.query generic

ecwyne:
Enables type information to be provided for query results.

```ts
client.query<IQueryResponse>(query).then(...)
```

### DIFF
--- a/src/types/Client.d.ts
+++ b/src/types/Client.d.ts
@@ -14,7 +14,7 @@ export interface ClientConfig {
 export default class Client {
   constructor(opts?: ClientConfig);
 
-  query(expr: Expr): Promise<object>;
+  query<T = object>(expr: Expr): Promise<T>;
   paginate(expr: Expr): PageHelper;
   ping(scope?: string, timeout?: number): Promise<string>;
 }


ecwyne:
Any chance you can take a look at this @erickpintor, should be fairly simple and would make a big difference in usability for those of use using typescript.

ecwyne:
@erickpintor `IQueryResponse` is actually just a placeholder for whatever is returned by the query (which you can define).

```typescript
client.query<string>(query).then(str => {...});
client.query<number>(query).then(num => {...});
client.query<Date>(query).then(date => {...});
client.query<{id: string, name: string}>(query).then(obj => {...});
```

It may be worth adding the following types I've been using though.


```typescript
interface IRef {
  id: string;
  class?: IRef
}

interface IFaunaInstance<T = {}> {
  ref: IRef;
  ts: number;
  data: T
}
```

Which can be used like this.

```typescript
type User = { username: string }
const query = q.Get(q.Ref(q.Class('users'), userId));
const user = await client.query<IFaunaInstance<User>>(query)
```

erickpintor:
This is pretty neat. Although, having an instance type being the default might get new users confused. `object` seems like a more universal default. I'll take this internally and let you know the outcome as soon as possible.

ecwyne:
Just wanted to be clear, this PR doesn't change the default behavior at all. The return type of client.query is still `Promise<object>` by default. I have just added the ability to _optionally_ provide a different type if you'd like.

erickpintor:
Sure. I understand that. We'll get it merged soon.

erickpintor:
Eric, this code has been merged and released into the version 2.6.1. Thank you very much for your contribution.

PR #175, opened 2019-01-26
merge duplicated condition

dbrrt:


danieltodonnell:
@dbrrt - Thanks for making this code change and thank you for your patience in releasing it.  We were planning on deploying it with some security updates we are publishing this week.

PR #179, opened 2019-08-01
Allow `null` as result of `q.If` expression

ptpaterson:
It has become useful to use 'null' as the result of a `q.If` expression.  The following is now possible:

```typescript
q.Let(
  {
    embeddedObject: q.Select(["data", "embeddedFieldName"], instance, null as any)
  },
  q.If(
    q.Equals([q.Var("embeddedObject"), null]),
    null,
    {
      /* q.Select stuff from the embeddedObject variable */
    }
  )
);
```


ptpaterson:
It likely doesn't make sense to add `null` straight to the `ExprVal` type.  Also, I don't know if this `null` is useful anywhere else, but it definitely is to me as `q.If` arguments.

marrony:
well, pretty much everything is nullable in FQL, so if you can, I think you can change the definition of `ExprArg`

BrunoQuaresma:
We have a ticket (https://faunadb.atlassian.net/browse/FE-33) to check this scenario and it will be fixed soon. Closing this until there.

ptpaterson:
Very cool, thanks!

BrunoQuaresma:
@danieltodonnell I think it is pretty ok to move on :). What do you think?

PR #180, opened 2019-08-21
Use agent with keepAlive option in Node.js

marszall87:
Subsequent queries are much faster, as requests are reusing the TLS connection.


fauna-ryan:
Thanks! This is now the default in node environments, configurable with the `keepAlive` option (docs: https://fauna.github.io/faunadb-js/Client.html#Client__anchor).

PR #188, opened 2019-10-22
CON-654: [js] Require http/https only in node environments

fauna-ryan:
[CON-654]

[CON-654]: https://faunadb.atlassian.net/browse/CON-654

### DIFF
--- a/src/Client.js
+++ b/src/Client.js
@@ -12,8 +12,6 @@ var json = require('./_json');
 var RequestResult = require('./RequestResult');
 var util = require('./_util');
 var PageHelper = require('./PageHelper');
-var http = require('http');
-var https = require('https');
 var parse = require('url-parse');
 
 /**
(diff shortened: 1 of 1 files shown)


src/Client.js
```diff
@@ -78,7 +76,9 @@ function Client(options) {
   this._lastSeen = null;
 
   if (isNodeEnv && opts.keepAlive) {
-    this._keepAliveEnabledAgent = new (isHttps ? https : http).Agent({ keepAlive: true });
+    this._keepAliveEnabledAgent = new (
+      isHttps ? require('https') : require('http')
+    ).Agent({ keepAlive: true });
```
BrunoQuaresma: I thought the dynamic requires was an issue. 🤔

fauna-ryan: not dynamic requires, just weird (maybe incorrect) syntax using `require` with a variable: https://github.com/fauna/faunadb-js/blob/2.8.1/src/Client.js#L73

no warnings/errors running this through webpack (e.g., `netlify-lambda`)

BrunoQuaresma: Ahh ok, cool!

PR #191, opened 2019-10-25
[CON-602]: Add ToDouble()/ToInteger()

fauna-ryan:
[CON-602]

[CON-602]: https://faunadb.atlassian.net/browse/CON-602

### DIFF
--- a/test/query_test.js
+++ b/test/query_test.js
@@ -1397,6 +1397,19 @@ describe('query', function () {
   it('to_array_and_to_object', function () {
     var obj = { k0: 10, k1: 20 };
     return assertQuery(query.ToObject(query.ToArray(obj)), obj);
+  })
+
+  it('to_double', function () {
+    var p1 = assertQuery(query.ToDouble(3.14), 3.14);
+    var p2 = assertQuery(query.ToString(query.ToDouble(10)), "10.0");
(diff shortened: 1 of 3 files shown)


test/query_test.js
```diff
@@ -1243,6 +1243,18 @@ describe('query', function () {
     return Promise.all([p1, p2]);
   });
 
+  it('to_double', function () {
+    var p1 = assertQuery(query.ToDouble(3.14), 3.14);
+    var p2 = assertQuery(query.ToDouble("3.14"), 3.14);
```
marrony: actually you are just testing if a double is converted to a double, add another test converting a "integer" to double.

```javascript
assertQuery(query.ToString(query.ToDouble(3.14)), "3.14");
assertQuery(query.ToString(query.ToDouble(10)), "10.0");
```

test/query_test.js
```diff
@@ -1243,6 +1243,18 @@ describe('query', function () {
...
+    return Promise.all([p1, p2]);
+  });
+
+  it('to_integer', function () {
+    var p1 = assertQuery(query.ToInteger(10.0), 10);
```
marrony: you could wrap you function on a `ToString(...)` function, that way you can assert that the output string doesn't have a fraction, like
```javascript
assertQuery(query.ToString(query.ToInteger(10.0)), "10");
```

PR #197, opened 2019-10-28
[CON-660]: Add type check functions

fauna-ryan:
[CON-660]

[CON-660]: https://faunadb.atlassian.net/browse/CON-660

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -5,5 +5,8 @@
     "es6": true,
     "jest": true,
     "node": true
+  },
+  "parserOptions": {
+    "ecmaVersion": 2018
   }
 }
--- a/src/query.js
+++ b/src/query.js
@@ -434,6 +434,307 @@ function IsNonEmpty(collection) {
   return new Expr({ is_nonempty: wrap(collection) })
 }
 
+// Type check functions
+
+/**
+ * Check if the expression is a number.
+ *
+ * @param {module:query~ExprArg} expr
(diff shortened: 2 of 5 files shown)


test/query_test.js
```diff
@@ -312,6 +324,136 @@ describe('query', function () {
...
+      query.CreateKey({ database: db, role: 'admin' })
+    ).then(pluckRef);
+    const tok = await serverClient.query(query.Login(doc, { password: 'sekret' }));
+    const role = await adminClient.query(query.CreateRole({
+      name: util.randomString('role_'),
+      membership: [],
+      privileges: [],
+    })).then(pluckRef);
+    // const sessionClient = new Client({ ...util.getCfg(), secret: tok.secret });
+    // const cred = await sessionClient.query(query.Paginate(query.Credentials()));
```
fauna-ryan: jvm:
```
    FaunaClient cli = serverClient.newSessionClient(tok.get(SECRET_FIELD));
    Value cred = cli.query(Get(Ref("credentials/self"))).get().at("ref");
```
i guess this is not the correct equivalent. any suggestions @marrony ?

marrony: no this is not, the only way to get credential is using the old ref model `Ref("credentials/self")`

fauna-ryan: Thanks. I had tried this previously without success, but it's working now.

PR #198, opened 2019-11-01
[CON-646]: Add prettier

fauna-ryan:
[CON-646]

closes #175 and #183

[CON-646]: https://faunadb.atlassian.net/browse/CON-646


BrunoQuaresma:
I guess these failed versions on CircleCI is because of newest FQL functions, right?

fauna-ryan:
> I guess these failed versions on CircleCI is because of newest FQL functions, right?

right

fauna-ryan:
> could you split this in two PR? pretty print the files is not related to the jest framework, so I think it would be better to commit those separate

sure: https://github.com/fauna/faunadb-js/pull/200

PR #202, opened 2019-11-06
[CON-650]: Add query options

fauna-ryan:
[CON-650]

[CON-650]: https://faunadb.atlassian.net/browse/CON-650


fauna-ryan:
> I missed a better description about what are the `query()` options and I think we have to update the typescript definition as well right?

Updated ts defs

BrunoQuaresma:
@fauna-ryan do you think makes sense to put a note about these options on README for non TS developers?

PR #203, opened 2019-11-07
[CON-693]: Add null to ExprArg

fauna-ryan:
[CON-693]
closes #179

[CON-693]: https://faunadb.atlassian.net/browse/CON-693

### DIFF
--- a/src/types/query.d.ts
+++ b/src/types/query.d.ts
@@ -1,6 +1,6 @@
 import Expr from './Expr'
 
-type ExprVal = Expr | string | number | boolean | object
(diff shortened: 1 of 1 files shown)


danieltodonnell:
I'm thinking that we want to make two additional types instead:

`ExprValNullable` and `ExprArgNullable`

and for now.. apply these to `If` only?

BrunoQuaresma:
I would say the best options is add `null` only in the `If` command but if we confirm that FQL expressions can also be null we should use null on ExprArg.

BrunoQuaresma:
Closing this one for now.

PR #210, opened 2019-12-17
Add custom headers

BrunoQuaresma:
Add the possibility to add custom headers on FaunaDB requests.


src/Client.js
```diff
@@ -219,6 +223,7 @@ Client.prototype._performRequest = function(
       'X-FaunaDB-API-Version': APIVersion,
       'X-Fauna-Driver': 'Javascript',
       'X-Last-Seen-Txn': this._lastSeen,
+      ...this._headers,
```
evbruno: is there a way for a user to override the previous headers set, such as `X-Fauna-Driver`, `X-Last-Seen-Txn`, etc. ?

BrunoQuaresma: Yes, it is. Should I remove this possibility?

evbruno: I know headers can be easily over-writen, but I guess: yes, remove it! :)
Or even better:
- some kind of filter to reject the ones we are already setting
- just exposing the business logic `setFaunaSource`. We have a setter for last seen txn (https://github.com/fauna/faunadb-js/blob/43665fbfd03bf3b9f9bc9de78ca7e6fb03312312/src/Client.js#L149)

BrunoQuaresma: 👍

Issue #207, opened 2019-11-28
Error trying to use in a Cloudflare Worker


yet-another-dev:
I'm seeing the following error while trying to use the driver in a [Cloudflare Worker](https://workers.cloudflare.com/): 

```
ReferenceError: XMLHttpRequest is not defined
```

This is making it impossible to use Fauna from Cloudflare Workers, unless there's some workaround that I haven't yet considered.

@aequasi [Sounds like](https://github.com/fauna/faunadb-js/issues/182#issuecomment-533427561) you're using Workers, are you seeing this issue?

Note: Originally opened this issue on the closed #182 but then realized it's likely best to create a new issue for it.

lolcoolkat:
Bump! Also getting this error. Any ideas on how we can fix this! :) Thanks!

lolcoolkat:
Update: Got it working on Cloudflare Workers :)

1) I had to remove the `cross-fetch` package (https://github.com/lolcoolkat/faunadb-js/commit/c2ad7a3e89ea3537888d8c9b7a1af12901aa8c91)

2) Also had to remove a little snippet at the bottom as it was giving an error as well (https://github.com/lolcoolkat/faunadb-js/commit/73ed874e796c1302e12f5d882c13c70edbdd3189)

I published the working version on NPM for the time being (so I could test it) https://www.npmjs.com/package/@snoculars/faunadb

Hopefully we can get these bugs fixed so I can use the official package again :D

BrunoQuaresma:
Hi @lolcoolkat thanks for the fix. 

1. We used `cross-fetch` because some environments was having compatibility issues. So we need to find another solution instead of just remove it.

2. I have to check what it is doing.

lolcoolkat:
@BrunoQuaresma Let me know if you have any time to look into a fix for this, or if you need me to work on this a bit? Would love to be able to use the official package soon at some point :D

BrunoQuaresma:
@lolcoolkat NodeJS does not support `fetch()` natively so we need to use some polyfill like `cross-fetch` to make driver works on both environments (browser and nodejs). Unfortunately, we didn't find out a polyfill that was not using XMLHttpRequest however we are brainstorming to see what we can do about this issue so please, let us know if you have any ideas on your mind.

eigilsagafos:
@BrunoQuaresma We use Axios in the browser and on our node backend. Have you considered using that?

BrunoQuaresma:
[...]

PR #214, opened 2020-01-08
Add fetch option

BrunoQuaresma:


BrunoQuaresma:
Related to: https://github.com/fauna/faunadb-js/issues/207

BrunoQuaresma:
@danieltodonnell the dynamic requires was solved. We are already doing that on https://github.com/fauna/faunadb-js/blob/1901da31f5087ff1dc552514e86a3d3906b34d0b/src/Client.js#L86

src/Client.js
```diff
@@ -77,6 +79,7 @@ function Client(options) {
   this._observer = opts.observer
   this._lastSeen = null
   this._headers = opts.headers
+  this._fetch = opts.fetch || require('cross-fetch')
```
marrony: is it common to use `require(...)` in the middle of the code?

BrunoQuaresma: It depends. It is used to avoid loading unnecessary stuff (on front-end side it reduces the bundle size).

lolcoolkat:
> Code-wise, this looks great, but we should NOT merge this into the main code-base until we have some major users test it out first. We had issues previously with dynamically required libraries, and I only want to `prerelease` this until we do thorough testing.
> NO MERGE YET.

I can do some debugging on this myself, and let you know if it works well or not!

BrunoQuaresma:
@lolcoolkat I would love that. Thanks!

lolcoolkat:
So I got it working inside my Cloudflare Worker by doing the following:
```
client: new faunadb.Client({
  secret,
  fetch: (req, init) => fetch(req, init),
})
```

**Note**: Passing `fetch` directly as a parameter gave me a **`TypeError: Illegal invocation`** error.

The only issue is that there seem's to be an issue with the following code, which gives me a **`TypeError: t.entries(...).forEach is not a function`** error. 
https://github.com/fauna/faunadb-js/blob/1901da31f5087ff1dc552514e86a3d3906b34d0b/src/Client.js#L189

Removing this line makes the FaunaDB client work fine again :)

BrunoQuaresma:
Thanks @lolcoolkat for the help on this.

1. `fetch: (req, init) => fetch(req, init)` does not sound right.
2. I'll check why the `.entries().forEach` is returning  `is not a function`.

mpavlinsky:
I am also trying to get this working on cloudflare workers so I happen to have some answers:
1. On cloudflare you can get the properly bound fetch with `fetch.bind(globalThis)`
2. The fetch headers entries are an 'Iterator' not an array-like (https://developer.mozilla.org/en-US/docs/Web/API/Headers/entries) so I think the fallback iteration needs to change to something like ```    var iter = responseHeaders.entries();
    var current;
    do {
      current = iter.next();
      if (current.value) {
        headers[current.value[0]] = current.value[1];
      }
    } while (!current.done);
```
(Or whatever the accepted way of iteration is in ES5, I assume we can't `for ... of`)

**Comment by lolcoolkat** [comment]
Thank you @mpavlinsky !! Doing the following:
```
return {
        client: new faunadb.Client({
          secret,
          fetch: fetch.bind(globalThis),
        }),
      }
``` 
seem's to do the trick! And for your #2 that info is helpful. Thanks!

**Comment by BrunoQuaresma** [comment]
@mpavlinsky do you know if there is any doc related to the `fetch()` API used by CloudFlare Worker? I'm trying to understand why it is diff of the "standard" web API and if there is another differences to add in the tests.

**Comment by mpavlinsky** [comment]
https://developers.cloudflare.com/workers/reference/apis/fetch/

Regarding the headers I think that is the behavior of the standard fetch. In my browser console `fetch('https://google.com', {mode: 'no-cors'}).then(r => console.log(r.headers.entries().forEach));` returns undefined.

**Comment by BrunoQuaresma** [comment]
@lolcoolkat @mpavlinsky can you check if the last commit fixes the `forEach()` issue?

**Comment by lolcoolkat** [comment]
> @lolcoolkat @mpavlinsky can you check if the last commit fixes the `forEach()` issue?

Works perfectly! Thanks for the fix :)

File: test/client.test.js [hunk 2]
```diff
@@ -111,6 +111,18 @@ describe('Client', () => {
 
     return expect(resultWithoutOptions).toEqual(resultWithOptions)
   })
+
+  test('uses custom fetch', async function() {
+    const fetch = jest.fn(() =>
+      Promise.resolve({
+        headers: new Set(),
```
mpavlinsky: I believe we want a test that exercises headers as an `Iterator` as is the case when using browser `fetch`

BrunoQuaresma: The `Iterator` is returned by the `entries()`. The `headers` options are type of `Headers[]` which have a similar API with the `Headers` object. I used `Set()` to avoid extra mocking.

BrunoQuaresma:
Beta version published on https://www.npmjs.com/package/faunadb/v/2.11.2-beta

PR #215, opened 2020-01-16
v2.10.3

BrunoQuaresma:


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 2.10.3
+
+- Fix If type
```
faunaee:
```suggestion
- Fix the types for `If` so that `null` is a valid result.
```

faunaee:
One other thought, is there a test that demonstrates success with a `null` result? Could be as simple as `q.If(true, null, null)`.

PR #221, opened 2020-01-24
FE-271 Fix Expr.toString() type def

BrunoQuaresma:
### Notes
[Jira Ticket](https://faunadb.atlassian.net/browse/FE-271)


src/types/Expr.d.ts
```diff
@@ -1,3 +1,5 @@
 export default class Expr {
   constructor(obj: object)
+
+  static toString(expr: Expr): string
```
fauna-ryanharris: @BrunoQuaresma Is this a utility method to convert Expressions into Strings?

BrunoQuaresma: Yes, it is.

PR #223, opened 2020-01-28
FE-227 | Improve arity error log information

fauna-ryanharris:
### Notes
[Jira Ticket](https://faunadb.atlassian.net/browse/FE-227)

This PR improves the error messages the user sees when calling a FQL method using the wrong number of arguments (i.e. incorrect arity).

I have updated the error logs to include:
1. The name of the FQL function that failed
2. A link to the Fauna docs for the function

Here's an example of how it'll look:
```
/Users/ryan/Code/faunadb-js/src/query.js:2828
    throw new errors.InvalidArity(min, max, args.length, callerFunc)
    ^
InvalidArity: Time function requires 1 argument(s) but 2 were given
For more info, see the docs: https://docs.fauna.com/fauna/current/api/fql/functions/time
```

### How to test
Refer to the `JS` section of this doc in Confluence: https://faunadb.atlassian.net/wiki/spaces/~254021588/pages/1008500759/Notes+Running+Fauna+drivers+from+repo+locally

1. Pull down and checkout this branch locally
2. Create a `test.js` at the project root
```javascript
var faunadb = require('./index')
var q = faunadb.query
 
  secret: 'secret',
  domain: 'localhost',
  port: 8443,
  scheme: 'http'
})
 
client.query(q.Time("1970-01-01T00:00:00+00:00"))
  .then((ret) => console.log(ret))
```
3. Copy the snippet in the article above
4. Update and run queries using different methods to confirm errors look correct


test/query.test.js
```diff
@@ -2357,77 +2357,71 @@ describe('query', () => {
...
-      Classes: [2, 'up to 1'],
-      Collections: [2, 'up to 1'],
-      Databases: [2, 'up to 1'],
-      Indexes: [2, 'up to 1'],
-      Functions: [2, 'up to 1'],
-      Roles: [2, 'up to 1'],
-      Keys: [2, 'up to 1'],
-      Tokens: [2, 'up to 1'],
-      Credentials: [2, 'up to 1'],
+      Ref: [3, 'from 1 to 2', 'Ref'],
```
fauna-ryanharris: @BrunoQuaresma @fauna-rodrigo I updated this test to reflect some of the changes I've made. However, the assertions we previously had did not really test all that much.

If y'all have any suggestions regarding what you would like our tests to cover here, I'm open to suggestions! 🙏

BrunoQuaresma: Maybe we could add an extra assertion to see if the documentation link is included. What do you think?

fauna-ryanharris: @BrunoQuaresma Sounds good to me! I'll add that and an assertion to confirm that the function name is in the first line of the log as well 👍

PR #225, opened 2020-02-12
FE-323

BrunoQuaresma:
### Notes
* Fix toExpr on fauna driver
  * Related to VSCode extension issue when the extension tries to parse a document as js and the content contains a string with double quotes inside.
* Fix typescript headers type
* Fix FaunaHTTPError typo on types

[Jira Ticket](https://faunadb.atlassian.net/browse/FE-323)

### How to test
* Using typescript, try to pass the headers params as config and check if there is any error. You can use `{ headers: { 'X-Fauna-Source': 'VSCode' }}`
* Create a collection like `{ data: '<p style=""></p>' }` and stringify it using `Expr.toString()` and see if there is any parser error.

### DIFF
--- a/test/query.test.js
+++ b/test/query.test.js
@@ -1076,16 +1076,12 @@ describe('query', () => {
         var documentRef = result.ref
         return client
           .query(query.Login(documentRef, { password: 'sekrit' }))
-          .then(function(result2) {
+          .then(async function(result2) {
             var secret = result2.secret
             var instanceClient = util.getClient({ secret: secret })
 
-            var self = new values.Ref(
-              'self',
-              new values.Ref('widgets', Native.COLLECTIONS)
-            )
             return instanceClient
-              .query(query.Select('ref', query.Get(self)))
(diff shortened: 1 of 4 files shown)


test/query.test.js
```diff
@@ -1076,16 +1076,12 @@ describe('query', () => {
         var documentRef = result.ref
         return client
           .query(query.Login(documentRef, { password: 'sekrit' }))
-          .then(function(result2) {
+          .then(async function(result2) {
```
danieltodonnell: Not ES5 compatible :(

BrunoQuaresma: It is only for tests. No production code at all.

BrunoQuaresma:
> All code (and tests) for now must be ES5 only.
It is already being used in other places of test code.

test/query.test.js
danieltodonnell: 👍

faunaee:
FYI: tests serve two purposes:
1) To automate validate of code.
2) Demonstrate how to use the code.

PR #230, opened 2020-03-04
v2.12.0

BrunoQuaresma:
Check the changelog to see the changes on this version.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,12 @@
+## 2.12.0
+
+- Add `headers` config on `Client`
+- Improve arity error log information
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,12 @@
+## 2.12.0
+
+- Add `headers` config on `Client`
```
faunaee: Will there be an update to `README.md` describing how to use the `headers` config?

BrunoQuaresma: I was thinking about that. Since it is more used by us internally I don't think we have to add a use case on README about this one. Opinions?

PR #232, opened 2020-03-06
Change release images for stable images

BrunoQuaresma:
### Notes
- This fixes the pipeline
[Jira Ticket](https://faunadb.atlassian.net/browse/FE-272)

### How to test
- Pipeline should be green

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,7 +8,7 @@ executors:
         type: integer
       version:
         type: enum
-        enum: ["release", "nightly"]
+        enum: ["stable", "nightly"]
     resource_class: large
     docker:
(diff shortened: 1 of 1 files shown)


BrunoQuaresma:
Even stable versions should be failing right now because the fix is only on nightly at this moment.

fauna-ryanharris:
@BrunoQuaresma So this is just changing the Docker image we use for tests from `release` to `stable`?

fauna-rodrigo:
I believe those types of changes can go directly to master

BrunoQuaresma:
@fauna-ryan yeap.

BrunoQuaresma:
@fauna-rodrigo I agree!

PR #237, opened 2020-03-20
Define constructor arguments of the RequestResult type as readonly properties

aleclarson:
Fixes the `RequestResult` type to match the runtime.

### DIFF
--- a/src/types/RequestResult.d.ts
+++ b/src/types/RequestResult.d.ts
@@ -1,5 +1,3 @@
-import Client from './Client'
-
 export default class RequestResult {
   constructor(
     method: string,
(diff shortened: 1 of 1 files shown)


BrunoQuaresma:
@aleclarson thanks for the contribution!

src/types/RequestResult.d.ts
```diff
@@ -15,5 +13,16 @@ export default class RequestResult {
     endTime: Date
   )
 
-  timeTaken: number
+  readonly method: string
+  readonly path: string
+  readonly query: object
```
danieltodonnell: `object?` maybe?
URL query parameters. Only set if `method` is "GET".

aleclarson: I just used the types from the `constructor` above shrug 🤷🏼‍♂️

PR #238, opened 2020-03-20
Expose "raw" property type on Expr class

aleclarson:


src/types/Expr.d.ts
```diff
@@ -1,5 +1,6 @@
 export default class Expr {
   constructor(obj: object)
+  readonly raw: object
```
marrony: That property is supposed to be private, actually you shouldn’t handling `Expr` objects, this is part of fauna DSL and is meant to be used with `client.query(...)` to be evaluated on the server side

aleclarson: I'm using it to check if an `Expr` is of a specific type, like this:
```ts
  q.LoadCollection = (e: string | Expr, opts?: object) => {
    if (typeof e === 'string') {
      e = q.Collection(e)
    }
    return q.Paginate('collection' in e.raw ? q.Documents(e) : e, opts)
  }

q.LoadCollection('foo')
```
I don't see the harm in doing so?

marrony: if you change this function to:
```javascript
  q.LoadCollection = (e: string | Expr, opts?: object) => {
    if (typeof e === 'string') {
      e = q.Documents(q.Collection(e))
    }
    return q.Paginate(e, opts)
  }

q.LoadCollection('foo')
```
you won't need to inspect `raw` property on `Expr`

aleclarson:
Just realized that duck-typing the `raw` object isn't foolproof, since `q.Var` queries are opaque in what they resolve to.

PR #252, opened 2020-03-31
feat: improved fql typings

aleclarson:
#### ⚠️ This is a work-in-progress. So far, only the functions that I use have improved types. All contributions are welcome! Just open a PR [here](https://github.com/aleclarson/faunadb-js) that targets the `types` branch. 🍻 

These functions are *possibly* finished (please review them):
- `q.Add`
- `q.Call`
- ~`q.Class`~ (deprecated)
- ~`q.Classes`~  (deprecated)
- `q.Collection`
- `q.Collections`
- `q.ContainsPath`
- `q.Create`
- `q.CreateCollection`
- `q.CreateFunction`
- `q.CreateIndex`
- `q.Delete`
- `q.Do`
- `q.Documents`
- `q.Equals`
- `q.Exists`
- `q.Filter`
- `q.Foreach`
- `q.Function`
- `q.Functions`
- `q.Get`
- `q.If`
- `q.Index`
- `q.Indexes`
- `q.IsEmpty`
- `q.Lambda`
- `q.Let`
- `q.Map`
- `q.Match`
- `q.Not`
- `q.Paginate`
- `q.Ref`
- `q.Select`
- `q.Time`
- `q.Update`

Unlisted functions haven't been touched yet.

When called, these functions *should* have their type parameters defined explicitly:
- `q.Index<MyDocument>(...)`
- `q.Collection<MyDocument>(...)`
- `q.CreateIndex<MyDocument>(...)`
- `q.CreateCollection<MyDocument>(...)`
- `q.Function<MyReturn>(...)`
- `q.CreateFunction<MyReturn>(...)`
- `q.Select<MyResult>(...)`

### DIFF
--- a/src/types/Expr.d.ts
+++ b/src/types/Expr.d.ts
@@ -122,4 +122,12 @@ export namespace Expr {
 
   /** Expression type for the `lambda` argument of `q.Filter` */
   export type Filter<T> = ExprVal<Lambda<[T], boolean>>
+
+  /** The expression type that can be mapped with `q.Map` */
+  export type Mappable<T = any> = Exclude<Iterable<T>, SetRef>
+
+  /** The expression type returned by `q.Map` */
+  export type MapResult<T extends Expr.Mappable, Out> = T extends Expr.Page
+    ? Expr.Page<Out>
+    : Expr<Out[]>
 }
--- a/src/types/Expr.d.ts
+++ b/src/types/Expr.d.ts
@@ -58,12 +58,6 @@ export interface Document<T extends object> {
   ts: number
 }
 
-export interface Page<T> {
-  data: T[]
-  before?: string
-  after?: string
-}
-
 export interface Index<T extends object, Meta extends object = any> {
   ref: Expr.IndexRef<T, Meta>
   ts: number
@@ -112,6 +106,13 @@ export namespace Expr {
     extends Ref<Function<Return, Meta>> {}
 
   export interface Time extends Expr<values.FaunaTime> {}
+
+  export interface Page<T>
+    extends Expr<{
+      data: T[]
+      before?: string
+      after?: string
+    }> {}
 
   /** Expression types that can be passed to `q.Map`, `q.Filter`, etc */
   export type Iterable<T> = ExprVal<T[]> | ArrayRef<T> | Page<T>
--- a/src/types/Expr.d.ts
+++ b/src/types/Expr.d.ts
@@ -88,11 +88,10 @@ export namespace Expr {
     private _refType: T
   }
 
-  /** @internal For type inference only. Please use `Expr.Iterable` instead. */
-  export interface ArrayRef<T = any> extends Ref<T[]> {}
-
-  /** Lazily evaluated set of refs */
-  export interface SetRef<T extends object = any> extends ArrayRef<Ref<T>> {}
+  export class SetRef<T = any> extends Expr<values.SetRef> {
+    // This prevents structural type equality.
+    private _type: 'SetRef' & T
+  }
 
   export interface DocumentRef<T extends object> extends Ref<Document<T>> {}
 
@@ -113,8 +112,8 @@ export namespace Expr {
     private _type: 'Page' & T
   }
 
-  /** Expression types that can be passed to `q.Map`, `q.Filter`, etc */
-  export type Iterable<T> = ExprVal<T[]> | ArrayRef<T> | Page<T>
+  /** The expression type for an iterable collection of values */
+  export type Iterable<T = any> = ExprVal<T[]> | SetRef<T> | Page<T>
 
   /** Expression type for the `path` argument of `q.Select` */
   export type KeyPath = ExprVal<(number | string)[]>
--- a/src/types/Expr.d.ts
+++ b/src/types/Expr.d.ts
@@ -77,8 +77,8 @@ export interface Index<T extends object> {
   permissions?: object
 }
 
-export interface Function<T extends object> {
-  ref: Expr.FunctionRef<T>
+export interface Function<Return> {
+  ref: Expr.FunctionRef<Return>
   ts: number
   name: string
   body: object
--- a/src/types/Expr.d.ts
+++ b/src/types/Expr.d.ts
@@ -104,35 +104,55 @@ export interface Function<Return, Meta extends object = any> {
 export namespace Expr {
   export class Ref<T = any> extends Expr<values.Ref> {
     // This prevents structural type equality.
-    private _type: 'Ref' & T
+    protected _type: 'Ref' & T
   }
 
   export class SetRef<T = any> extends Expr<values.SetRef> {
     // This prevents structural type equality.
-    private _type: 'SetRef' & T
+    protected _type: 'SetRef' & T
   }
 
-  export interface DocumentRef<T extends object> extends Ref<Document<T>> {}
+  export class DocumentRef<T extends object> extends Ref<Document<T>> {
+    // This prevents structural type equality.
+    protected _data: T
+  }
 
-  export interface CollectionRef<T extends object, Meta extends object = any>
-    extends Ref<Collection<T, Meta>> {}
+  export class CollectionRef<
+    T extends object,
+    Meta extends object = any
+  > extends Ref<Collection<T, Meta>> {
+    // This prevents structural type equality.
+    protected _data: T
+    protected _meta: Meta
+  }
 
-  export interface IndexRef<T extends object, Meta extends object = any>
-    extends Ref<Index<T, Meta>> {}
+  export class IndexRef<
+    T extends object,
+    Meta extends object = any
+  > extends Ref<Index<T, Meta>> {
+    // This prevents structural type equality.
+    protected _data: T
+    protected _meta: Meta
+  }
 
-  export interface FunctionRef<Return, Meta extends object = any>
-    extends Ref<Function<Return, Meta>> {}
+  export class FunctionRef<Return, Meta extends object = any> extends Ref<
+    Function<Return, Meta>
+  > {
+    // This prevents structural type equality.
+    protected _return: Return
+    protected _meta: Meta
+  }
 
   /** The expression type for a timestamp (nanosecond precision) */
   export class Time extends Expr<values.FaunaTime> {
     // This prevents structural type equality.
-    private _type: 'Time'
+    protected _type: 'Time'
   }
 
   /** The expression type for a page from a paginated set returned by `q.Paginate` */
   export class Page<T = any> extends Expr<values.Page<T>> {
     // This prevents structural type equality.
-    private _type: 'Page' & T
+    protected _type: 'Page' & T
   }
 
   /** The expression type for an iterable collection of values */
--- a/src/types/Expr.d.ts
+++ b/src/types/Expr.d.ts
@@ -1,6 +1,103 @@
-export default class Expr {
+import { values } from './values'
+
+export default Expr
+
+export class Expr<T = any> {
   constructor(obj: object)
 
   readonly _isFaunaExpr?: boolean
   static toString(expr: Expr): string
+
+  // This prevents structural type equality with empty objects.
+  private _exprType: T
+}
+
+/** Materialize an `Expr` type into its result type. */
+export type Materialize<T> = T extends Expr<infer U>
+  ? U
+  : T extends ReadonlyArray<infer U>
+  ? (U[] extends T
+      ? ReadonlyArray<Materialize<U>>
+      : { [P in keyof T]: Materialize<T[P]> })
+  : T
+
+/** Add support for `Expr` types to any type. */
+export type ExprVal<T = unknown> =
+  | Expr<Exclude<T, Expr>>
+  | (T extends Expr | Lambda
+      ? T
+      : T extends ReadonlyArray<infer U>
+      ? (U[] extends T
+          ? ReadonlyArray<ExprVal<U>>
+          : { [P in keyof T]: ExprVal<T[P]> })
+      : T extends object
+      ? { [P in keyof T]: ExprVal<T[P]> }
+      : T)
+
+export type Lambda<In extends any[] = any[], Out = any> = (
+  ...args: In
+) => Expr<Out>
+
+export interface Collection<T> {
+  ref: Expr.CollectionRef<T>
+  ts: number
+  name: string
+  data?: object
+  permissions?: object
+  history_days: number | null
+  ttl_days?: number
+}
+
+export interface Document<T> {
+  data: T
+  ref: Expr.DocumentRef<T>
+  ts: number
+}
+
+export interface Page<T> {
+  data: T[]
+  before?: string
+  after?: string
+}
+
+export interface Index<T> {
+  ref: Expr.IndexRef<T>
+  ts: number
+  name: string
+  data?: object
+  source: Expr.CollectionRef<any> | any[]
+  partitions: number
+  active: boolean
+  serialized?: boolean
+  unique?: boolean
+  terms?: any[]
+  values?: any[]
+  permissions?: object
+}
+
+export interface Function<T> {
+  ref: Expr.FunctionRef<T>
+  ts: number
+  name: string
+  body: object
+  role?: any
+  data?: object
+}
+
+export namespace Expr {
+  export abstract class Ref<T> extends Expr<values.Ref> {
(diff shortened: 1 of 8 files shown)


aleclarson:
An unavoidable weakness is lack of type guards. So narrowing an `Expr<string | number>` type to just `Expr<number>` using `q.IsNumber` won't be possible in the near future (if ever). TypeScript isn't concerned with lazy evaluation modes of programming. :(

src/types/Expr.d.ts
```diff
@@ -1,5 +1,109 @@
...
+  data?: object
+}
+
+export namespace Expr {
+  export abstract class Ref<T extends object = any> extends Expr<values.Ref> {
+    // This prevents structural type equality with empty objects.
+    private _refType: T
+  }
+
+  export interface SetRef<T extends object = any> extends Ref<Ref<T>[]> {}
```
marrony, line 92: `SetRef` is not a ref of ref, based on your work, it would be better represented as
```
  export abstract class SetRef<T extends object = any> extends Expr<values.SetRef> {
     private _refType: T
   }
```

src/types/query.d.ts
```diff
@@ -1,28 +1,109 @@
...
+  export function Call<T>(ref: Expr.FunctionRef<T>, ...args: any[]): Expr<T>
+
+  // TODO
   export function Query(lambda: ExprArg | Lambda): Expr
-  export function Map(collection: ExprArg, lambda_expr: ExprArg | Lambda): Expr
+
+  export function Map<T, Out>(
+    collection: Expr.Iterable<T>,
+    lambda_expr: ExprVal<Lambda<[T extends Expr ? T : Expr<T>], Out>>
+  ): Expr<Out[]>
```
marrony, line 104: the output wound't be `Expr<Iterable<Out>>`?

src/types/query.d.ts
```diff
@@ -66,20 +154,48 @@ export module query {
...
+
+  // TODO
   export function Update(ref: ExprArg, params: ExprArg): Expr
   export function Replace(ref: ExprArg, params: ExprArg): Expr
-  export function Delete(ref: ExprArg): Expr
+
+  export function Delete(
+    ref: Expr.Ref<any>
+  ): Expr<{
+    database: string
```
marrony, line 192: I don't get why the output of `Delete` would always have `database`field, this only happens if you are deleing a database

src/types/query.d.ts
aleclarson: depends on what you expect `client.query(q.Map(...))` to resolve to.

if the mapping function returns `Expr<number>`, I would expect `Promise<number[]>` to be returned by `client.query`

src/types/Expr.d.ts
```diff
@@ -1,5 +1,118 @@
...
+
+export type Lambda<In extends any[] = any[], Out = any> = (
+  ...args: { [P in keyof In]: ToExpr<In[P]> }
+) => Expr<Out>
+
+export interface Collection<T extends object> {
+  ref: Expr.CollectionRef<T>
+  ts: number
+  name: string
+  data?: object
```
marrony, line 46: you can add a second parameter to that interface to represent the metadata to enforce this type instead of use `object`

src/types/query.d.ts
aleclarson: good to know! 👍 

i'm not using the return value of `Delete` atm, so someone else will need to make the change

src/types/Expr.d.ts
aleclarson: yup, i'm about to push such a change to support that :)

src/types/query.d.ts
```diff
@@ -66,20 +154,48 @@ export module query {
   export function IsCredentials(expr: ExprArg): Expr
   export function IsRole(expr: ExprArg): Expr
 
-  export function Get(ref: ExprArg, ts?: ExprArg): Expr
+  export function Get<T extends object>(
+    ref: Expr.Ref<T>,
```
marrony, line 158: `Get` can either receive a `Ref` or a `SetRef`

aleclarson: correct, this is how it is already, but will need to be updated when `SetRef` is changed as you suggested earlier

src/types/query.d.ts
```diff
@@ -173,29 +315,66 @@ export module query {
...
   export function Database(name: ExprArg, scope?: ExprArg): Expr
-  export function Index(name: ExprArg, scope?: ExprArg): Expr
-  export function Class(name: ExprArg, scope?: ExprArg): Expr
-  export function Collection(name: ExprArg, scope?: ExprArg): Expr
-  export function Function(name: ExprArg, scope?: ExprArg): Expr
+
+  // TODO: "scope" argument
+  export function Index<T extends object>(
+    name: ExprVal<string>,
+    scope?: ExprArg
```
marrony, line 326: `scope` is a `DatabaseRef`

src/types/Expr.d.ts
aleclarson: ok, will update 👍

src/types/query.d.ts
```diff
@@ -253,6 +434,10 @@ export module query {
   export function ToMillis(expr: ExprArg): Expr
   export function ToMicros(expr: ExprArg): Expr
 
+  // TODO
   export function MoveDatabase(from: ExprArg, to: ExprArg): Expr
-  export function Documents(collection: ExprArg): Expr
+
+  export function Documents<T extends object>(
+    collection: Expr.CollectionRef<T>
+  ): Expr.SetRef<Document<T>>
```
marrony, line 442: `Documents()` function won't return a set of `Document`, but instead will return a set of `Ref`s

src/types/query.d.ts
```diff
@@ -173,29 +315,66 @@ export module query {
...
+  // TODO
   export function Role(name: ExprArg, scope?: ExprArg): Expr
   export function Databases(scope?: ExprArg): Expr
-  export function Classes(scope?: ExprArg): Expr
-  export function Collections(scope?: ExprArg): Expr
-  export function Indexes(scope?: ExprArg): Expr
-  export function Functions(scope?: ExprArg): Expr
+
+  // TODO: "scope" argument
+  export function Collections(scope?: ExprArg): Expr.SetRef<Collection<any>>
```
marrony, line 346: it will return `Expr.SetRef<CollectionRef<any>>`

src/types/query.d.ts
aleclarson: that's true, but the `Expr.SetRef` type makes declaring such more convenient by wrapping its type parameter with `Expr.Ref`.

so that means `Expr.SetRef<Document<T>>` resolves to `Expr.Ref<Array<Expr.Ref<Document<T>>>>`

of course, that will change slightly when i update `Expr.SetRef` as suggested

src/types/query.d.ts
```diff
@@ -173,29 +315,66 @@ export module query {
...
+
+  // TODO: "scope" argument
+  export function Indexes(scope?: ExprArg): Expr.SetRef<Index<any>>
```
marrony, line 349: `Expr.SetRef<IndexRef<any>>`

src/types/query.d.ts
```diff
@@ -173,29 +315,66 @@ export module query {
...
+
+  // TODO: "scope" argument
+  export function Functions(scope?: ExprArg): Expr.SetRef<Function<any>>
```
marrony, line 352: `Expr.SetRef<FunctionRef<any>>`

src/types/query.d.ts
aleclarson: right. it currently does do that ;)

src/types/query.d.ts
aleclarson: yup, that's what it does (try it out with `type Foo = Expr.SetRef<Index<any>>`)

src/types/query.d.ts
marrony: `q.Map` would never return anything different than an Iterable

src/types/query.d.ts
aleclarson: i'm not using fql with database refs currently, so this will remain untyped until someone contributes

src/types/query.d.ts
aleclarson: Not sure what you mean. When doesn't `q.Map` return an array?

```js
const result = await db.query(q.Map([1], x => q.Add(x, 1)))
result // => [ 2 ]
```

marrony: I said it always will return an iterable, if you pass an array it will return an array, if you pass a page it will return a page. Map doesnt support sets yet but should in the future

marrony: Pages
```javascript
> Paginate(Databases(), {size: 4})
{
  after: [ Database("blah") ],
  data: [
    Database("blah_renamed"),
    Database("db1"),
    Database("db2"),
    Database("test_driver")
  ]
}
```

Map+Pages
```javascript
> Map(Paginate(Databases(), {size: 4}), Lambda(ref => Get(ref)))
{
  after: [ Database("blah") ],
  data: [
    {
      ref: Database("blah_renamed"),
      ts: 1578425890150000,
      priority: 1,
      name: 'blah_renamed'
    },
    {
      ref: Database("db1"),
      ts: 1583508872680000,
      name: 'db1',
      data: { gql: { ts: Time("2020-03-06T15:34:32.296894Z") } }
    },
    { ref: Database("db2"), ts: 1563914425250000, name: 'db2' },
    {
      ref: Database("test_driver"),
      ts: 1569504687220000,
      name: 'test_driver'
    }
  ]
}
```

aleclarson: Fixed 👍

src/types/Expr.d.ts
```diff
@@ -1,5 +1,124 @@
...
+  export interface FunctionRef<Return, Meta extends object = any>
+    extends Ref<Function<Return, Meta>> {}
+
+  export interface Time extends Expr<values.FaunaTime> {}
+
+  export interface Page<T>
+    extends Expr<{
+      data: T[]
+      before?: string
+      after?: string
```
aleclarson, line 113: What should cursors be typed as?

src/types/query.d.ts
```diff
@@ -1,46 +1,177 @@
...
+    arg4: ExprVal<T>
+  ): ToExpr<T>
+  export function Do<T>(
+    arg1: ExprVal,
+    arg2: ExprVal,
+    arg3: ExprVal,
+    arg4: ExprVal,
+    arg5: ExprVal<T>
+  ): ToExpr<T>
+  export function Do<T>(...args: ExprVal[]): ToExpr<T>
```
aleclarson, line 120: The `q.Do` function needs to return the last expression type, which is hard because TypeScript has poor support for getting the last type in a varargs type. For now, I've declared an overload for between 1 and 5 arguments. Any more and you need the return type must be explicit.

src/types/Expr.d.ts
marrony: usually the same type `T`, but things change a bit when you use Paginate+Map, because the mapped page will have a new type different than the initial, but the cursor type will always be the same, so I would add a second argument to the interface, one for data types and another for cursors.

src/types/query.d.ts
marrony: Conceptually your `SetRef` definition is not defining what a set is, so this is making functions like this giving the wrong impression to the users that `Documents(...)` will return a set of documents, but instead this will return a set of references, I would like to see the definition of that function something like:
```typescript
  export function Documents<T extends object>(
     collection: Expr.CollectionRef<T>
   ): Expr.SetRef<DocumentRef<T>>
```
then update the `SetRef` accordingly

aleclarson: Doesn't a `SetRef` always contain `Ref` values only?

benjumanji: Nope. If you page an index with a non-default values set you'll get a set over those values, which are not constrained to be refs.

src/types/query.d.ts
```diff
@@ -88,29 +251,53 @@ export module query {
...
 
+  // TODO
   export function Singleton(ref: ExprArg): Expr
   export function Events(ref_set: ExprArg): Expr
-  export function Match(index: ExprArg, ...terms: ExprArg[]): Expr
+
+  export function Match<T extends object>(
+    index: Expr.IndexRef<T>,
+    ...terms: any[]
+  ): Expr.SetRef<Document<T>>
```
marrony, line 283: this is wrong as well, it should return `SetRef<T>`, because when materialized that set won't return documents, but it will depends on the index values

src/types/query.d.ts
marrony: no, the `Match(...)` function is an example (which is wrong by your definition) that returns a SetRef, but when materialized by `Paginate` will transform into pages of type `T`, depending on the index definition

src/types/query.d.ts
aleclarson: yeah, I've assumed `IndexRef` to contain documents only for now, because I'm not using custom values yet. 

Contribution welcome!

steida:
Wow, I just discovered this PR. Great work.

steida:
@aleclarson Did you think about using `JsonObject` instead of object type? It could help people to not accidentally trying to store Date instance or other invalid types. I can send PR.

```ts
type Json = string | number | boolean | null | JsonObject | JsonArray;

interface JsonObject {
  [property: string]: Json;
}

interface JsonArray extends Array<Json> {}
```

aleclarson:
@steida Sure!

steida:
Sure what? :-) Send PR? Is it safe to replace all object types with JsonObject? I did it in a fork and it seemed to work. I can't figure out why it should not.

aleclarson:
@steida Idk, I'll give it a look when you open the PR. 👍

steida:
@aleclarson Done.

src/types/query.d.ts
```diff
@@ -1,46 +1,172 @@
...
+  export function Lambda<In extends any[], Out>(
+    f: (...args: In) => Out
+  ): Expr<Lambda<In, Out>>
+
+  export function Lambda<In extends any[], Out>(
+    var_name: ExprArg<string>,
+    expr: ExprArg<Out>
+  ): Expr<Lambda<In, Out>>
+
+  export function Call<T>(ref: Expr.FunctionRef<T>, ...args: any[]): Expr<T>
```
marrony, line 134: `ref` can be either a function reference or string with the function name

src/types/query.d.ts
```diff
@@ -1,46 +1,172 @@
...
-  export function Lambda(f: Lambda): Expr
-  export function Lambda(var_name: ExprArg, expr: ExprArg): Expr
-  export function Call(ref: ExprArg, ...args: ExprArg[]): Expr
+
+  export function Lambda<In extends any[], Out>(
+    f: (...args: In) => Out
+  ): Expr<Lambda<In, Out>>
+
+  export function Lambda<In extends any[], Out>(
+    var_name: ExprArg<string>,
```
marrony, line 130: In that form, `var_name` is either a single string or an array of strings, never an expression

aleclarson:
Thanks to @steida for being the first contributor! 🎉 🙏 ❤️

BrunoQuaresma:
@aleclarson thanks for your hard work. There are some open conversations from @marrony, so let us know if you need something from us.

aleclarson:
I just pushed a fix (4ea60a8bbd539aa77a369743e874f36d5f084902) so that `Ref` types (eg: `DocumentRef<T>`) extend the `Ref` class, which lets you pass `new values.Ref('xxx')` as a query expression. Also, `Ref` types are now preserved when mapped by the `Materialize<T>` type.

And I rebased onto `master` branch, too. 😁 

@BrunoQuaresma FYI: I don't plan on finishing this PR up on my own, unless I end up needing every single FaunaDB feature before others contribute (which is unlikely).

aleclarson:
@steida Btw, one shortcoming of `JsonObject` is that it forces `interface` types to extend `JsonObject` before you can use them in `db.collection<MyInterface>(...)` (and similar) calls. Not sure if I hate it enough to go back to `object` or not. 😆

winghouchan:
Thanks for all your work on this @aleclarson ❤️

I see the last commits and conversations were 17+ months ago. What's the current status of this PR? What's needed for a merge? Is there a possibility of breaking down this PR and incrementally delivering it so folks can see value earlier and it may make reviewing it easier? Are there things that folks can do to contribute to moving this forward?

PR #261, opened 2020-04-10
FE-471 | Configure browserify to output ES module

fauna-ryanharris:
### Notes
[Jira Ticket](https://faunadb.atlassian.net/browse/FE-471)

In this PR, I did the following to configure `browserify` to also output a ES Module:
1. Added the following packages: `babelify`, `@babel/core`, `@babel/preset-env` and `@babel/preset-react`
2. Update `browserify` script in `package.json` to output a `.es.js` file transformed by `babelify`

### How to test
I tested this locally by going into my `faunadb-js` repo and:
1. Running `yarn link`
2. Setting up a CRA and running `yarn link faunadb`
3. Added the following to my `App.js`
```javascript
import faunadb, { query } from "faunadb";
```
4. Logged out `faunadb` and `query`. Got the following:
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,6 +25,7 @@
     "require": "dist/faunadb.js"
   },
   "type": "module",
+  "module": "dist/faunadb.es.js",
   "scripts": {
     "doc": "jsdoc -c ./jsdoc.json",
     "browserify": "browserify index.js --standalone faunadb -o dist/faunadb.js | browserify index.js -o dist/faunadb.es.js -t [ babelify --presets [ @babel/preset-env @babel/preset-react ] ]",
--- a/package.json
+++ b/package.json
@@ -20,7 +20,9 @@
     "dist/"
   ],
   "main": "index.js",
-  "module": "dist/faunadb.es.js",
+  "exports": {
+    "browser": "dist/faunadb.es.js"
+  },
   "scripts": {
     "doc": "jsdoc -c ./jsdoc.json",
     "browserify": "browserify index.js --standalone faunadb -o dist/faunadb.js | browserify index.js -o dist/faunadb.es.js -t [ babelify --presets [ @babel/preset-env @babel/preset-react ] ]",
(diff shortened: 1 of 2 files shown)


fauna-ryanharris:
@BrunoQuaresma @danieltodonnell @fauna-rodrigo Only question I have is if we need to update our `browserify-min` script 🤔 I don't think so, but I'll do a little more research.

steida:
@fauna-ryanharris I think there should be `"module": "...es.js"` in package.json file. You can test it by importing only one module, then compare compiled size. In Next.js, `yarn build` is enough.

fauna-ryanharris:
> @fauna-ryanharris I think there should be `"module": "...es.js"` in package.json file. You can test tree-shaking by importing only one module, then compare compiled sizes. In Next.js, `yarn build` is enough.

@steida Thanks for the note -- I'll double check that! I think you're right because I've had to do that in previous side projects using `microbundle` 👍

package.json
```diff
@@ -20,9 +20,15 @@
     "dist/"
   ],
   "main": "index.js",
+  "exports": {
+    "import": "dist/faunadb.es.js",
+    "require": "dist/faunadb.js"
+  },
+  "type": "module",
+  "module": "dist/faunadb.es.js",
```
fauna-ryanharris, line 28: @steida I added the `module` field for future support though [this StackOverflow post](https://stackoverflow.com/a/42817320) says it is just a proposal at this point.

To be safe, I also added the `exports` objects as per these Node docs: https://nodejs.org/api/esm.html#esm_conditional_exports

fauna-ryanharris:
I tested this with `yarn link` in both a Create React App and a Gatsby project. I also ran a local Node script using the CJS syntax. All three worked 👍

BrunoQuaresma:
We just need to be sure why the night tests are failing. My concern is that maybe this change is breaking some environments / use cases.

PR #264, opened 2020-04-23
FE-484 | Add queryTimeout to Client

fauna-ryanharris:
### Notes
[Jira Ticket](https://faunadb.atlassian.net/browse/FE-484)

**To Do:**

- [x] ~Add some tests to cover these changes~

The work in this PR does the following:
1. Adds a `queryTimeout()` function to the `Client.prototype`, which allows users to set the `_queryTimeout` field on the Client instance

```javascript
const client = new faunadb.Client({
  secret: 'MY_KEY'
})
client.queryTimeout(200)
```

2. Allows a user to pass a `queryTimeout` value to the `client.query()` function call, which will override `Client._queryTimeout` if it also has a value

```javascript
const data = client.query(q.Paginate(q.Collections()), { queryTimeout: 100 })
```

### How to test
To test locally, I did the following:

1. In the local `faunadb-js` directory, run `yarn link`
2. In the local `console` repo, run `yarn link faunadb` 
3. At the root of `console`, create a `test.js`
4. Import `faunadb` and try to use the new functionality. Test by running `node test.js`.
5. In `faunadb-js`, add some `console.log`s to `Client.prototype._performRequest` to inspect the outgoing headers

My script looked like this:
```javascript
const faunadb = require('faunadb'),
  q = faunadb.query

const client = new faunadb.Client({
  secret: 'MY_KEY'
})
client.queryTimeout(200)

const data = client.query(q.Paginate(q.Collections()), { queryTimeout: 100 })
data.then(res => console.log(res)).catch(err => console.log(err))
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,16 +153,22 @@ const client = new faunadb.Client({
 })
 ```
 
-On the other hand, using the client's `queryTimeout` dictates how long FaunaDB will process the request on the server before timing out if it hasn't finished running the operation. This can be done in two different ways:
+On the other hand, using the client's `queryTimeout` dictates how long FaunaDB will process the request on the server before timing out if it hasn't finished running the operation. This can be done in three different ways:
 
 ```javascript
-// 1. Setting the value on the client instance
+// 1. Setting the value when instantiating a new client
+const client = new faunadb.Client({
+  queryTimeout: 2000,
+  secret: 'YOUR_FAUNADB_SECRET',
+})
+
+// 2. Setting the value on the client instance
 const client = new faunadb.Client({
   secret: 'YOUR_FAUNADB_SECRET',
 })
 client.queryTimeout(200)
 
-// 2. Specifying the value per-query
+// 3. Specifying the value per-query
 var data = client.query(q.Paginate(q.Collections()), {
   queryTimeout: 100,
 })
--- a/README.md
+++ b/README.md
@@ -144,7 +144,7 @@ The client can be configured to handle timeouts in two different ways:
 1. Add a `timeout` field to the `options` block when instantiating the client
 2. By setting a `queryTimeout` on the client (or passing the value to the client's `.query()` method directly)
 
-The first option (i.e. `timeout`) represents a HTTP timeout on the client side. Represented in milliseconds, the client will wait the specified period before timing out if it has yet to receive data.
+The first option (i.e. `timeout`) represents a HTTP timeout on the client side. Represented in milliseconds, the client will wait the specified period before timing out if it has yet to receive a response.
 
 ```javascript
 const client = new faunadb.Client({
--- a/README.md
+++ b/README.md
@@ -144,7 +144,7 @@ The client can be configured to handle timeouts in two different ways:
 1. Add a `timeout` field to the `options` block when instantiating the client
 2. By setting a `queryTimeout` on the client (or passing the value to the client's `.query()` method directly)
 
-The first option (i.e. `timeout`) represents a HTTP timeout on the client side. Represented in milliseconds, the client will wait the specified period before timing out if it has yet to receive a response.
+The first option (i.e. `timeout`) represents a HTTP timeout on the client side. Defined in milliseconds, the client will wait the specified period before timing out if it has yet to receive a response.
 
 ```javascript
 const client = new faunadb.Client({
--- a/README.md
+++ b/README.md
@@ -153,7 +153,7 @@ const client = new faunadb.Client({
 })
 ```
 
-On the other hand, using the client's `queryTimeout` dictates how long FaunaDB will process the request on the server-side before timing out if it hasn't finished it's lookup and/or returned data. This can be done in two different ways:
+On the other hand, using the client's `queryTimeout` dictates how long FaunaDB will process the request on the server before timing out if it hasn't finished running the operation. This can be done in two different ways:
 
 ```javascript
 // 1. Setting the value on the client instance
--- a/README.md
+++ b/README.md
@@ -153,7 +153,7 @@ const client = new faunadb.Client({
 })
 ```
 
-On the other hand, using the client's `queryTimeout` dictates how long FaunaDB will process the request on the server before timing out if it hasn't finished running the operation. This can be done in three different ways:
+On the other hand, using the client's `queryTimeout` dictates how long FaunaDB will process the request on the server before timing out if it hasn't finished running the operation. This can be done in two different ways:
 
 ```javascript
 // 1. Setting the value when instantiating a new client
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -137,9 +137,42 @@ helper
...
 more information on the pagination helper.
 
+#### Timeouts
+
+The client can be configured to handle timeouts in two different ways:
+
+1. Add a `timeout` field to the `options` block when instantiating the client
+2. By setting a `queryTimeout` on the client (or passing the value to the client's `.query()` method directly)
+
+The first option (i.e. `timeout`) represents a HTTP timeout on the client side. Represented in milliseconds, the client will wait the specified period before timing out if it has yet to receive data.
```
nathaniel-may, line 147:
```suggestion
The first option (i.e. `timeout`) represents a HTTP timeout on the client side. Represented in milliseconds, the client will wait the specified period before timing out if it has yet to receive a response.
```

README.md
```diff
@@ -137,9 +137,42 @@ helper
...
+
+```javascript
+const client = new faunadb.Client({
+  secret: 'YOUR_FAUNADB_SECRET',
+  timeout: 100,
+})
+```
+
+On the other hand, using the client's `queryTimeout` dictates how long FaunaDB will process the request on the server-side before timing out if it hasn't finished it's lookup and/or returned data. This can be done in two different ways:
```
nathaniel-may, line 156:
```suggestion
On the other hand, using the client's `queryTimeout` dictates how long FaunaDB will process the request on the server before timing out if it hasn't finished running the operation. This can be done in two different ways:
```

nathaniel-may:
Fantastic detailed changes to the README. A++++++

src/Client.js
```diff
@@ -104,6 +105,14 @@ function Client(options) {
 Client.prototype.query = function(expression, options) {
```
marrony, line 105: non related to this PR, but I didn't know `.query` function receives an `options` argument

fauna-ryanharris: @marrony Same here! #TIL 🚀

test/client.test.js
```diff
@@ -123,6 +123,38 @@ describe('Client', () => {
...
     expect(fetch).toBeCalled()
   })
+
+  test('set queryTimeout on client instance', () => {
+    client.queryTimeout(1000)
+    expect(client._queryTimeout).toEqual(1000)
+
+    client.queryTimeout(2000)
+    expect(client._queryTimeout).toEqual(2000)
+  })
```
BrunoQuaresma, line 133: Test private fields are a bad test pattern since it is testing the implementation and not the interface.

test/client.test.js
```diff
@@ -123,6 +123,38 @@ describe('Client', () => {
...
+
+  test.only('properly sets query timeout header', async () => {
```
BrunoQuaresma, line 135: You should remove the `.only`.

test/client.test.js
```diff
@@ -123,6 +123,38 @@ describe('Client', () => {
...
+    })
+
+    const testClient = util.getClient({
+      fetch,
+    })
+
+    await testClient.query(query.Paginate(query.Collections()), {
+      queryTimeout: 2000,
+    })
+  })
```
BrunoQuaresma, line 157: Maybe are you missing an `expect`? I think you can check something like `expect(fetch).wasCalledWith(...)`

fauna-ryanharris: @BrunoQuaresma I have an `expect` on line 137 so I can make sure our `X-Query-Timeout` header is going out properly with the request. Do you think that is enough to cover this functionality?

src/Client.js
```diff
@@ -227,6 +244,7 @@ Client.prototype._performRequest = function(
       'X-FaunaDB-API-Version': APIVersion,
       'X-Fauna-Driver': 'Javascript',
       'X-Last-Seen-Txn': this._lastSeen,
+      'X-Query-Timeout': queryTimeout,
```
danieltodonnell, line 247: If no `queryTimeout` is specified, would we want this header to emit?

BrunoQuaresma:
> Really liked the documentation, guys.
> good job
> 
> One question: I could not find it so easily, but did u write a test for making sure the queryTimeout has priority over the regular timeout option ?

Maybe this is related to what your are asking? https://github.com/fauna/faunadb-js/pull/264/files#diff-0183c8c134bf806b1b2c33d41990222cR197

src/Client.js
BrunoQuaresma: It is ok to have. No problems are caused passing this header as empty.

fauna-ryanharris: @danieltodonnell Now, you can pass the queryTimeout value when instantiating the Client. If no value is passed, it defaults to null. Then, the `removeNullAndUndefinedValues` call on line 241 removes any headers that are `undefined` or `null`.

On line 234, we check the `options` object passed to `.query()` to see if a `queryTimeout` has been passed. If so, we use that value instead of the Client's `this._queryTimeout` 👍

README.md
```diff
@@ -137,9 +137,48 @@ helper
...
+const client = new faunadb.Client({
+  queryTimeout: 2000,
+  secret: 'YOUR_FAUNADB_SECRET',
+})
+
+// 2. Setting the value on the client instance
+const client = new faunadb.Client({
+  secret: 'YOUR_FAUNADB_SECRET',
+})
+client.queryTimeout(200)
```
marrony, line 169: usually a I don't like mutate the object state, if I'm using the same client to issue multiple queries and change this property in one part of the code, this will have side effects in other parts of code not related, my opinion is to remove this option and leave it only on constructor and on query method

src/Client.js
nathaniel-may: Great clarification regarding `removeNullAndUndefinedValues`. Probably warrants a comment at this point in the code too.

README.md
nathaniel-may: Personally, I also prefer having immutable clients too, but I think that's relatively uncommon for other similar JS clients. Before this change was the client completely immutable already?

src/Client.js
danieltodonnell: Thanks for all the due diligence everyone!

README.md
marrony: besides this setter https://github.com/fauna/faunadb-js/blob/master/src/Client.js#L152 that's meant to be used to sync last txn time (internally only in my option) there are no other setter.

nathaniel-may: Ok so in that case I'm inclined to not introduce mutability into the client with this change.

Here are two options I can think of:
1. remove the `.queryTimeout()` setter.
2. make the `.queryTimeout()` method return a new client with the set parameter.

fauna-ryanharris: @nathaniel-may I discussed this with the team and we are in favor of #1. If users end up asking for something like #2, we can always add it later. What do you think?

nathaniel-may: Sounds good to me. Thanks for raising the issue, @marrony.

fauna-ryanharris:
@nathaniel-may @marrony I implemented the changes you asked for regarding the `.queryTimeout()` method. You can now only set the `queryTimeout` on the Client when instantiating it. You can also pass a value to `.query()` to specify on a per-query basis.

I've updated README to reflect these changes. Please let me know if you'd like any more detail/context in the docs 👍

PR #265, opened 2020-05-04
new serialization proposal

marrony:
This is a proposal to change values serialization, so, internally FaunaClient utilize a custom function to escape values by adding `@` tag, bug leave the native json serialization as is.

### DIFF
--- a/src/_json.js
+++ b/src/_json.js
@@ -1,14 +1,64 @@
 'use strict'
 
 var values = require('./values')
+var Expr = require('./Expr')
+
+function faunaTypes(key, value) {
+  if (key === '') {
+    value = this[key]
+  }
+
+  if (
+    key !== '' &&
+    typeof value === 'undefined' &&
+    typeof this !== 'undefined'
+  ) {
+    value = this[key]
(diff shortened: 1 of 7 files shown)


BrunoQuaresma:
@marrony is this a break change?

nathaniel-may:
could we see some code examples of what users could do after the change?

aleclarson:
+1

Currently, the `Ref` type is lying about its properties, but this PR fixes that.

Is this a breaking change?

src/_json.js
```diff
@@ -1,14 +1,64 @@
 'use strict'
 
 var values = require('./values')
+var Expr = require('./Expr')
+
+function faunaTypes(key, value) {
+  if (key === '') {
+    value = this[key]
```
aleclarson, line 8: Why is this part necessary?

aleclarson:
Is `@obj` an artifact of a past version? I noticed it's decoded by the parser, but never encoded by the replacer.

doom-weaver:
closing as this is 2 years old.

PR #268, opened 2020-05-06
Fix a bug with instanceof conditions

gahabeen:
### Notes
The checks with `instanceof` fail to properly handle query expressions in certain cases (when there are more than one faunadb instance for ex).

### How to test
Havn't been able to test it. If someone else can (or I'll try again later).

For reference, here is where comes from the picked solution (#3): https://stackoverflow.com/a/41592087. And the reference of the thread in slack where we talked about it: https://fauna-community.slack.com/archives/CLMPZHJ0G/p1588784775089500.


gahabeen:
@BrunoQuaresma @ryancharris It was painful to test/fix in my environment but that's it, it works!

It a really minor push. Only adding a couple of readonly props to custom types and then checking for them in the tests.

Hope you guys can push that soon 🙏

gahabeen:
Added `obj !== null` tests to avoid failed reads of properties on null values.

test.js
```diff
@@ -0,0 +1,12 @@
+const fauna = require('./index')
+const { query: q } = fauna
+const db = new fauna.Client({
+  secret: 'fnADrL7VfSACAhtIplPXl_C-iX20yk6hQbReCxAI',
+})
+
+;(async () => {
+  // let res = await db.query(q.Let({ super: true }, q.Var('super')))
```
fauna-ryanharris, line 8: @gahabeen If they're not needed, can you please remove these comments? Thanks!

gahabeen: Damned, these have slipped in the push. My bad 😅
I'll remove them.

gahabeen: Done 👌

fauna-ryanharris: Thank you! 🙏

src/query.js
```diff
@@ -2920,6 +2928,9 @@ function wrap(obj) {
   } else if (obj instanceof Uint8Array || obj instanceof ArrayBuffer) {
     return new values.Bytes(obj)
   } else if (typeof obj === 'object') {
+    if (obj.constructor.name === 'Expr') {
+      console.log(obj.constructor.name, Object.keys(obj), obj)
```
BrunoQuaresma, line 2932: Could we remove this?

gahabeen: Dang, good catch. I'll remove it in a minute.

gahabeen:
> 
> 
> @gahabeen Looking at the code it looks hard to understand what is happening and what is the goal 🤔. Do you think there is a way to be more clear in the code to express what you are trying to achieve?

The idea is to be able to do a proper check of an object being the instance of a `Ref` or an `Expr`. In some cases, @fauna-brecht as myself ended up having with more than one instance of `faunadb` implying tests failing when `faunadb1.Expr({ }) instanceof faunadb2.Expr !== true`.

Hence why going further and adding a property to the prototype chain to avoid wrongly failed tests.
It doesn't change the previous behavior as it only adds an alternative match that check if the value is a non-null object and check if the property exists.

If you see any way to make it better, I'm all for it. 
We could for ex refactor it in a function as follows:
```js
function checkInstanceWithProperty(obj, prop) {
   return (typeof obj === 'object' && obj !== null && Boolean(obj[prop]))
}
```

BrunoQuaresma:
great, thanks @gahabeen for the explanation. More questions: What is the use case of you need to have two faunadb instances? (I just trying to understand the full scenario).

gahabeen:
@BrunoQuaresma I believe is has to do with the test environment (using symlinks). For me it happened having 3 local libs (biota's modules) and when using 2 (local) libs in the 3rd one I'ld get this error.

It's not an error that many people will get. But eventually if they have faunadb installed in a couple of local packages and they want to use them one in the other, it will fail.

fauna-brecht:
To further elaborate, there is no use-case persé, it's just unavoidable if your complex gets more complicated. For example, I want to reuse fauna queries for frontend, backend or user-defined functions. Which means I separate my fauna queries from backend or frontend and have the following folders. 

- backend
- fauna-queries
- frontend

The only way to make this work and make sure that node or any other compilation does not include multiple versions of the fauna library in the dependencies is to make sure to put 
peerDependencies in fauna-queries instead of a direct dependency. 

However, I also have tests in fauna-queries which won't run properly since peerDependencies cant be picked up by tests, not even if I run it from another package that provides the fauna dependency. To run these I have to put the fauna library in devDependencies again and that somehow breaks it again making node and/or create-react-app include multiple versions of the fauna library. 

In essence, it's just how the javascript eco-system works which will eventually result in users having multiple versions of the library and that will make sure that instanceof fails as in this StackOverflow issue: https://stackoverflow.com/questions/41587865/using-instanceof-on-objects-created-with-constructors-from-deep-npm-dependenci/41592087#41592087

That means that everyone that writes a library to be used by other users should apparently avoid instanceOf.

BrunoQuaresma:
@gahabeen do you think is it possible to abstract some of these `(typeof obj === 'object' && obj !== null && Boolean(path._isFaunaValue))` to a variable or a function with a name? like `isFaunaInstance` or `isValidExpr`? Maybe abstract these conditionals and put a name on it can make life easier to read. What do you think?

gahabeen:
> 
> 
> > @gahabeen Looking at the code it looks hard to understand what is happening and what is the goal 🤔. Do you think there is a way to be more clear in the code to express what you are trying to achieve?
> 
> The idea is to be able to do a proper check of an object being the instance of a `Ref` or an `Expr`. In some cases, @fauna-brecht as myself ended up having more than one instance of `faunadb` implying tests failing when `faunadb1.Expr({ }) instanceof faunadb2.Expr !== true`.
> 
> _(This error would result in deep wrapping of values in my queries. It was nasty and took me a day to find out what was happening 😅)_
> 
> Hence why going further and adding a property to the prototype chain to avoid wrongly failed tests.
> It doesn't change the previous behavior as it only adds an alternative match that check if the value is a non-null object and check if the property exists.
> 
> If you see any way to make it better, I'm all for it.
> We could for ex refactor it in a function as follows:
> 
> ```js
> function checkInstanceWithProperty(obj, prop) {
>    return (typeof obj === 'object' && obj !== null && Boolean(obj[prop]))
> }
> ```

@BrunoQuaresma What about (my earlier offer) using a function to abstract the fuzziness of the test? If you like it I can add it.

We will always need to be checking for `object`, `null` and then the differentiating parameter (as `_isFaunaRef` for the `Ref` constructor or `_isFaunaValue` for the `value` constructor) else it will fail (because the tested value can be anything).

BrunoQuaresma:
@gahabeen looks good! Sorry for miss that.

gahabeen:
Added the `checkInstanceHasProperty` fn :)

gahabeen:
Fixed the missing wrapped tests.

fauna-brecht:
Works fine for Fwitter and effectively solves my sharing code problems

gahabeen:
Any ETA on this?

BrunoQuaresma:
@gahabeen let's merge this right now but to release a new version I think only next week. Sorry for delayed this.

BrunoQuaresma:
@gahabeen also, could you please update the package-lock? After that I'll merge it directly.

gahabeen:
Sounds good @BrunoQuaresma. Done the merge.

PR #281, opened 2020-05-26
FE-540 | Mis-cased FQL function names

fauna-ryanharris:
### Notes
[Jira Ticket](https://faunadb.atlassian.net/browse/FE-540)

The following FQL function names were improperly cased in the UDF form due to the logic below:

- `ContainsStr`
- `ContainsStrRegex`
- `StartsWith`
- `EndsWith`
- `RegexEscape`
- `FindStrRegex`
- `ReplaceStr`
- `FindStr`

```javascript
  fn = fn
    .split('_')
    .map(function(str) {
      return str.charAt(0).toUpperCase() + str.slice(1)
    })
    .join('')
```

The issue is that not all FQL function names come across the wire `in_snake_case`, so the logic above only capitalizes the first letter. Here's an example:

```javascript
// Wire protocol
replacestrregex

// Expected result
ReplaceStrRegex

// Actual result
Replacestrregex
```

### How to test
1. Create a new UDF using one of the functions listed above
2. Save it
3. Confirm the form does not re-case the function name anymore

### Screenshots

### DIFF
--- a/src/Expr.js
+++ b/src/Expr.js
@@ -39,6 +39,14 @@ var varArgsFunctions = [
   'And',
   'Or',
 ]
+
+// FQL function names come across the wire as all lowercase letters
+// (like the key of this object). Not all are properly snake-cased
+// on the Core side, which causes improper capitalizations.
+//
+// JS Driver patch: https://faunadb.atlassian.net/browse/FE-540
+// Core update: https://faunadb.atlassian.net/browse/ENG-2110
+
 var specialCases = {
   containsstr: 'ContainsStr',
   containsstrregex: 'ContainsStrRegex',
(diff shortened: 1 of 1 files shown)


fauna-ryanharris:
> LGTM. Great to have these "special cases" already in place 👏

@BrunoQuaresma Agreed. Any other function names that have this issue in the future can just be added to the list too 👍

fauna-ryanharris:
> I'd just add a comment to indicate what `SpecialCases` is used for and why it's necessary.
> 
> Also, I deeply appreciate that the list is alphabetical.

@nathaniel-may I'll add that note 👍 And, of course, I **had** to re-order them for my own OCD 😆

PR #282, opened 2020-05-27
FE-625: Fix Paginate/Match rendering

marrony:


BrunoQuaresma:
@marrony what was the issue?

marrony:
```javascript
> Get(Function("clean_ref_history"))
{
  ref: Function("clean_ref_history"),
  ts: 1588262941410000,
  name: 'clean_ref_history',
  body: Query(Lambda(["ref", "after"], Foreach(Paginate(Events(Var("ref")), Var("after"), 1024), Lambda("e", Remove(Select("document", Var("e")), Select("ts", Var("e")), Select("action", Var("e")))))))
}
```
this is the problem

BrunoQuaresma:
@marrony sorry but I really didn't see the issue....

marrony:
have you tried to execute the function body in the shell?

evbruno:
hmmmm... the current create function throws this error:

```
[InvalidArity: Paginate function requires from 1 to 2 argument(s) but 3 were given
For more info, see the docs: https://docs.fauna.com/fauna/current/api/fql/functions/paginate] {
  name: 'InvalidArity',
  message: 'Paginate function requires from 1 to 2 argument(s) but 3 were given\n' +
    'For more info, see the docs: https://docs.fauna.com/fauna/current/api/fql/functions/paginate',
  min: 1,
  max: 2,
  actual: 3
```

evbruno:
>  sorry but I really didn't see the issue....

when in doubt, look at the test sources... so, if you revert the core changes, and run the tests, you can glimpse at the fix...

eg: before the fix,

the query 
`Query(q.Lambda('_', q.Paginate(q.Match(q.Index('idx')))))` 
will render 
`Query(Lambda("_", Paginate(Match(Index("idx"), []))))` 
instead of 
`'Query(Lambda("_", Paginate(Match(Index("idx")))))'`

BrunoQuaresma:
@evbruno thanks for the explanation, now I can see!

Issue #287, opened 2020-06-07
Wrong timeout docs  in readme


steida:
It's defined in seconds, not milliseconds.

BrunoQuaresma:
Taking a look on it, thanks.

fauna-ryanharris:
https://github.com/fauna/faunadb-js/pull/291

PR #291, opened 2020-06-09
FE-645 | docs: fix timeout unit in README

fauna-ryanharris:
### Notes
[Jira Ticket](https://faunadb.atlassian.net/browse/FE-645)

In our README, the unit of measure for the Client's `timeout` option was listed as milliseconds but is actually expecting a value in terms of seconds.

**Note:** Instead of this change, we could also make the `timeout` and `queryTimeout` options both be consistent and measure in milliseconds.


fauna-ryanharris:
closes #287

PR #296, opened 2020-06-12
FE-684 | Add ContainsValue

fauna-ryanharris:
### Notes
[Jira Ticket](https://faunadb.atlassian.net/browse/FE-648)

This PR adds a new function to the JS driver called `ContainsValue`. This functionality has existed in the Core for a few weeks, but has yet to be added here.

### How to test
To properly test this, you will need to make sure you're using the latest version of our `nightly` Docker image.

1. You can test this functionality by running our local test suite: `npm run test`
2. You could `yarn link` your local JS Driver repo to your local Console repo and use the webshell

### DIFF
--- a/src/Client.js
+++ b/src/Client.js
@@ -1,6 +1,6 @@
 'use strict'
 
-var APIVersion = '2.7'
+var APIVersion = '3'
 
 var btoa = require('btoa-lite')
 var errors = require('./errors')
--- a/src/query.js
+++ b/src/query.js
@@ -1959,6 +1959,20 @@ function Contains(path, _in) {
 /**
  * See the [docs](https://app.fauna.com/documentation/reference/queryapi#miscellaneous-functions).
  *
+ * @param {module:query~ExprArg} value
+ *   Represent the value we want to search for.
+ * @param {module:query~ExprArg} in
+ *   An object we will search for the value passed in.
+ * @return {Expr}
+ */
+function ContainsValue(value, _in) {
+  arity.exact(2, arguments, ContainsValue.name)
+  return new Expr({ contains_value: wrap(value), in: wrap(_in) })
+}
+
+/**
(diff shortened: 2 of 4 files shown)


BrunoQuaresma:
@ryancharris is this stills a WIP?

fauna-ryanharris:
> @ryancharris is this stills a WIP?

@BrunoQuaresma It's still WIP for now. I need to get more clarification re: how this function is expected to work, so I can add any additional test coverage that's needed.

fauna-ryanharris:
@nathaniel-may @marrony Please let me know if there are any use cases I may have missed and I'll add more coverage 👍

nathaniel-may:
Honestly, this looks pretty good to me. Coverage here regarding function behavior doesn't need to be 100% since we have that covered in core tests. It just needs to be sent and received correctly and these test cases cover that.

marrony:
> Honestly, this looks pretty good to me. Coverage here regarding function behavior doesn't need to be 100% since we have that covered in core tests. It just needs to be sent and received correctly and these test cases cover that.

fauna-ryanharris:
@nathaniel-may @marrony I'm going to add some more tests coverage, especially now that I have a bit better of an understanding re: FQL. I'll re-request your reviews once I have added some more cases 👍

nathaniel-may:
> I don't agree 100%. Core has its own "driver" and as far as I can tell, it sometimes behaves differently than the drivers. One example is the `Min/Max/And/Or` functions. Core always send an array of values, but drivers has a tendency to replace the array case it has only one element.

You're 100% right. I just see these as two different pieces of functionality to test. 1- the evaluation fo the FQL expression (very little coverage required by the drivers), and 2- the conversion of that expression to a `local type`. If a driver has an extra conversion that does need to be tested.

PR #297, opened 2020-06-15
FE-485 | Add Reverse function

fauna-ryanharris:
### Notes
[Jira Ticket](https://faunadb.atlassian.net/browse/FE-485)

**To Do:**
- [x] Add multi-Page test

This PR adds a new `Reverse` function that takes a Set, Array or Page and returns the object's items in reverse order.

### How to test
Run `npm run test` in your local repo to make sure the test suites pass

### DIFF
--- a/src/Client.js
+++ b/src/Client.js
@@ -1,6 +1,6 @@
 'use strict'
 
-var APIVersion = '2.7'
+var APIVersion = '3'
 
 var btoa = require('btoa-lite')
 var errors = require('./errors')
--- a/test/query.test.js
+++ b/test/query.test.js
@@ -2363,12 +2363,18 @@ describe('query', () => {
     const reverseDBs = await client.query(
       query.Reverse(query.Paginate(query.Indexes()))
     )
-
-    console.log('databases', databases)
-    console.log('reverseDBs', reverseDBs)
+    const documentsSet = await client.query(
+      query.Paginate(query.Documents(query.Collection('widgets')))
+    )
+    const reverseDocumentsSet = await client.query(
+      query.Reverse(
+        query.Paginate(query.Documents(query.Collection('widgets')))
+      )
+    )
 
     assertQuery(query.Reverse(numArray), numArray.reverse())
     expect(databases.data).toEqual(reverseDBs.data.reverse())
+    expect(documentsSet.data).toEqual(reverseDocumentsSet.data.reverse())
   })
 
   // Check arity of all query functions
--- a/test/query.test.js
+++ b/test/query.test.js
@@ -2357,7 +2357,7 @@ describe('query', () => {
     expect(results.data).toHaveLength(20)
   })
 
-  test.only('reverse', async () => {
+  test('reverse', async () => {
     const numArray = [1, 2, 3]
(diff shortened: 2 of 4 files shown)


src/Client.js
```diff
@@ -1,6 +1,6 @@
 'use strict'
 
-var APIVersion = '2.7'
+var APIVersion = '3'
```
nathaniel-may, line 3: In the future, API version bumps can be their own PRs.

test/query.test.js
```diff
@@ -1906,17 +1906,19 @@ describe('query', () => {
   test('and', () => {
     var p1 = assertQuery(query.And(true, true, false), false)
     var p2 = assertQuery(query.And(true, true, true), true)
-    var p3 = assertQuery(query.And(true), true)
-    var p4 = assertQuery(query.And(false), false)
-    return Promise.all([p1, p2, p3, p4])
+    // var p3 = assertQuery(query.And(true), true)
+    // var p4 = assertQuery(query.And(false), false)
+    // return Promise.all([p1, p2, p3, p4])
```
nathaniel-may, line 1911: What's the commented code?

fauna-ryanharris: @nathaniel-may I uncommented these. Originally, I needed to comment them out because we had removed `And` and `Or`'s ability to take a single argument. @marrony added that back to Core, so now the tests will pass without them being commented out.

src/Client.js
fauna-ryanharris: @nathaniel-may In order to get our tests to pass, I need to use the `3` header or else the Core throws an error 🙏

nathaniel-may: right. what I mean is that we can issue a PR with this line only then start feature work. It's a minor point.

fauna-ryanharris: @nathaniel-may Ah, okay, I misunderstood. Apologies!

test/query.test.js
```diff
@@ -2393,6 +2393,79 @@ describe('query', () => {
...
+        query.Paginate(query.Match(query.Index('widgets_by_n'), 100), {
+          size: 2,
+          after: [anderson.ref],
+        })
+      )
+    )
+
+    // Set
+    const documentsSet = await client.query(
+      query.Paginate(query.Documents(query.Collection('widgets')))
```
marrony, line 2455: if you are calling paginate, then it's not a set, it's a page

test/query.test.js
```diff
@@ -2393,6 +2393,79 @@ describe('query', () => {
...
+    )
+    const reverseDocumentsSet = await client.query(
+      query.Reverse(
+        query.Paginate(query.Documents(query.Collection('widgets')))
```
marrony, line 2459: if you are calling paginate first, then you are reversing a page, not the set

fauna-ryanharris: @marrony Thanks for the note 🙏 Instead, should I do `Paginate(Reverse(Documents(Collection(...))))`?

marrony: yes

fauna-ryanharris:
@marrony I updated the test for Sets. Lemme know if it looks good to you 🙏

PR #299, opened 2020-06-19
FE-665 | Add AccessProvider document type and tests

fauna-ryanharris:
### Notes
[Jira Ticket](https://faunadb.atlassian.net/browse/FE-665)

**Additional specs:** https://faunadb.atlassian.net/browse/ENG-1941

This PR does the following:
1. Adds a `AccessProvider` document type
2. Adds definition of `AccessProvider` to TS types file
3. Implements logic for JSON parsing
4. Adds tests to cover instantiation and JSON parsing/encoding of `AccessProvider`

### How to test
Run our local test suite: `npm run test`

### DIFF
--- a/src/_json.js
+++ b/src/_json.js
@@ -21,6 +21,17 @@ function json_parse(_, val) {
     return val
   } else if ('@ref' in val) {
     var ref = val['@ref']
+    var isAccessProvider = ref.name && ref.issuer && ref.jwks_url
+
+    if (isAccessProvider) {
+      return new values.AccessProvider(
+        ref.name,
+        ref.issuer,
+        ref.jwks_url,
+        ref.allowed_roles,
+        ref.allowed_collections
+      )
+    }
 
     if (!('collection' in ref) && !('database' in ref)) {
       return values.Native.fromName(ref['id'])
@@ -42,15 +53,6 @@ function json_parse(_, val) {
     return new values.Bytes(val['@bytes'])
   } else if ('@query' in val) {
     return new values.Query(val['@query'])
-  } else if ('@access_provider' in val) {
-    var ref = val['@access_provider']
-    return new values.AccessProvider(
-      ref.name,
-      ref.issuer,
-      ref.jwks_url,
-      ref.allowed_roles,
-      ref.allowed_collections
-    )
   } else {
     return val
   }
--- a/src/_json.js
+++ b/src/_json.js
@@ -42,6 +42,15 @@ function json_parse(_, val) {
     return new values.Bytes(val['@bytes'])
   } else if ('@query' in val) {
     return new values.Query(val['@query'])
+  } else if ('@access_provider' in val) {
+    var ref = val['@access_provider']
+    return new values.AccessProvider(
+      ref.name,
+      ref.issuer,
+      ref.jwks_url,
+      ref.allowed_roles,
+      ref.allowed_collections
+    )
   } else {
     return val
   }
--- a/src/_json.js
+++ b/src/_json.js
@@ -21,17 +21,6 @@ function json_parse(_, val) {
     return val
   } else if ('@ref' in val) {
     var ref = val['@ref']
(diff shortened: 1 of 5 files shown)


src/_json.js
```diff
@@ -42,6 +42,15 @@ function json_parse(_, val) {
     return new values.Bytes(val['@bytes'])
   } else if ('@query' in val) {
     return new values.Query(val['@query'])
+  } else if ('@access_provider' in val) {
+    var ref = val['@access_provider']
```
marrony, line 46: This is not correct, `AccessProviders` are just a new reference type for drivers point of view

fauna-ryanharris: @marrony Thanks. Should this be using `@ref` instead?

marrony: you should do this as part of `@ref` parsing

fauna-ryanharris:
@marrony I adjusted the parsing based on your note. Let me know what you think 👍

src/values.js
```diff
@@ -364,6 +364,52 @@ function wrapToString(type, fn) {
...
+ * @param {string} name A valid schema name
+ * @param {string} issuer A unique string
+ * @param {string} jwks_url A valid HTTPS URL
+ * @param {Array<Ref>} [allowed_roles=[]] A list of Role refs
+ * @param {Array<Ref>} [allowed_collections=[]] A list of user-defined Collection refs
+ *
+ * @extends module:values~Value
+ * @constructor
+ */
+function AccessProvider(
```
freels, line 377: Adding a specific document type for AccessProvider seems like more specificity than necessary at least compared to how the driver currently handles documents of other types. (Collection, Index, Database, Function, etc)

src/values.js
```diff
@@ -172,6 +173,7 @@ var Native = {
   FUNCTIONS: new Ref('functions'),
   ROLES: new Ref('roles'),
   KEYS: new Ref('keys'),
+  ACCESS_PROVIDERS: new Ref('accessProviders'),
```
marrony, line 176: `access_providers`

src/values.js
```diff
@@ -188,6 +190,8 @@ Native.fromName = function(name) {
       return Native.ROLES
     case 'keys':
       return Native.KEYS
+    case 'accessProviders':
```
marrony, line 193: `access_providers`

src/values.js
```diff
@@ -123,6 +123,7 @@ wrapToString(Ref, function() {
     indexes: 'Index',
     functions: 'Function',
     roles: 'Role',
+    accessProviders: 'AccessProvider',
```
marrony, line 126: `access_providers: 'AccessProvider'`

fauna-ryanharris: @marrony That's what I had originally, however, when snake-casing it, the pretty print results in `Access_provider` and fails our tests.

marrony: there is no sneak case conversion in this place, this is used to convert from schema name to the function name

fauna-ryanharris: I had all three instances as `access_provider` in `value.js` before changing them to `accessProviders`. Still got that result 🤔

marrony: two different things going on here:

1. Rendering of `AccessProvider("blah")`
  https://github.com/fauna/faunadb-js/blob/master/src/values.js#L134
  to fix this you just need to add `access_providers: 'AccessProvider'` on `constructors` object

2. Rendering of `AccessProviders([db])`
  https://github.com/fauna/faunadb-js/blob/master/src/values.js#L131
  this would involve more change, but for the sake of making the tests pass you can just add 
  `      if (ref.id === 'access_providers') return 'AccessProviders(' + db + ')'` before the line 131

fauna-ryanharris: @marrony Thanks for the notes. I was just doing some investigation and was about to ping you about item 2 since I traced the issue there as well 👍

marrony: for the item 2 you can do some fancy/generic stuff like here https://github.com/fauna/faunadb-js/blob/master/src/Expr.js#L150 or a simple `if` should be enough

PR #300, opened 2020-06-19
FE-663 | Add CreateAccessProvider function

fauna-ryanharris:
### Notes
[Jira Ticket](https://faunadb.atlassian.net/browse/FE-663)
**Additional specs:** https://faunadb.atlassian.net/browse/DRV-124

This PR adds the `CreateAccessProvider` function 👍 

### How to test
~Tests currently won't pass. I [need to get clarification re: how to test this function ](https://github.com/fauna/faunadb-js/pull/300#pullrequestreview-434210888)or remove the `create_access_provider` test block in `query.test.js`~

Run `npm run test`

### DIFF
--- a/src/query.js
+++ b/src/query.js
@@ -1024,6 +1024,23 @@ function CreateRole(params) {
   return new Expr({ create_role: wrap(params) })
 }
 
+/**
+ * See the [docs](https://app.fauna.com/documentation/reference/queryapi#write-functions).
+ *
+ * @param {module:query~ExprArg} params
+ *   An object of parameters used to create a new access provider.
+ *     - name: A valid schema name
+ *     - issuer: A unique string
+ *     - jwks_url: A valid HTTPS URL
+ *     - allowed_roles: An optional list of Role refs
+ *     - allowed_collections: An optional list of user-defined Collection refs
+ * @return {Expr}
+ */
+function CreateAccessProvider(params) {
+  arity.exact(1, arguments, CreateAccessProvider.name)
+  return new Expr({ create_access_provider: wrap(params) })
+}
+
 // Sets
 
 /**
@@ -3074,6 +3091,7 @@ module.exports = {
   CreateKey: CreateKey,
   CreateFunction: CreateFunction,
   CreateRole: CreateRole,
+  CreateAccessProvider: CreateAccessProvider,
   Singleton: Singleton,
   Events: Events,
   Match: Match,
--- a/src/query.js
+++ b/src/query.js
@@ -1031,7 +1031,7 @@ function CreateRole(params) {
  *   An object of parameters used to create a new access provider.
  *     - name: A valid schema name
  *     - issuer: A unique string
- *     - jwks_url: A valid HTTPS URL
+ *     - jwks_uri: A valid HTTPS URI
  *     - allowed_roles: An optional list of Role refs
  *     - allowed_collections: An optional list of user-defined Collection refs
  * @return {Expr}
--- a/test/query.test.js
+++ b/test/query.test.js
@@ -848,6 +848,18 @@ describe('query', () => {
     })
   })
 
+  test.only('create access provider', async () => {
+    const provider = await adminClient.query(
+      query.CreateAccessProvider({
+        name: 'my_provider',
+        issuer: 'user-123',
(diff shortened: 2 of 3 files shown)


src/types/values.d.ts
```diff
@@ -55,6 +55,22 @@ export module values {
     constructor(value: object)
   }
 
+  export class AccessProvider extends Value {
```
fauna-ryanharris, line 58: This file is part of #299 diff

src/values.js
```diff
@@ -364,6 +364,52 @@ function wrapToString(type, fn) {
   }
 }
 
+/**
```
fauna-ryanharris, line 367: This files is part of the diff from #299

test/query.test.js
```diff
@@ -2393,7 +2406,7 @@ describe('query', () => {
     expect(results.data).toHaveLength(20)
   })
 
-  test.only('reverse', async () => {
+  test('reverse', async () => {
```
fauna-ryanharris, line 2409: The changes in this file are both from #299

test/values.test.js
```diff
@@ -12,7 +12,8 @@ var FaunaDate = values.FaunaDate,
   Ref = values.Ref,
   SetRef = values.SetRef,
   Bytes = values.Bytes,
-  Query = values.Query
+  Query = values.Query,
```
fauna-ryanharris, line 15: This file is from #299

test/query.test.js
```diff
@@ -848,6 +848,19 @@ describe('query', () => {
     })
   })
 
+  test.only('create_access_provider', async () => {
+    // TODO: Define or remove this test
+    const accessProvider = await client.query(
```
fauna-ryanharris, line 853: @nathaniel-may @marrony Any advice re: how I can go about testing this? I'm assuming I need to authenticate or point this to a different URL cause right now I get a ` PermissionDenied: permission denied` error.

marrony: you need to use a admin key to create access providers

fauna-ryanharris: @marrony Thanks. I figured it was something like that. Would something like this work?

```javascript
    const accessProvider = await client.query(
      query.CreateAccessProvider({
        name: 'my-provider',
        issuer: 'ryan',
        jwks_url: 'https://db.fauna.com',
      }), { secret: ADMIN_KEY}
    )
```

test/query.test.js
```diff
@@ -848,6 +848,18 @@ describe('query', () => {
     })
   })
 
+  test.only('create access provider', async () => {
+    const provider = await adminClient.query(
```
fauna-ryanharris, line 852: @marrony I am getting a ` BadRequest: validation failed` error here. I'm using the `adminClient` now since it has an admin key baked in 🤔

marrony:
```
[
  {
    position: [ 'create_access_provider' ],
    code: 'validation failed',
    description: 'document data is not valid.',
    failures: [
      {
        field: [ 'jwks_uri' ],
        code: 'value required',
        description: 'Required value not found.'
      }
    ]
  }
]
```

fauna-ryanharris: @marrony Looks like on the Core side it's called `jwks_uri` whereas the ticket I implemented had it listed as `jwks_url`. I'll updated the field and confirm that fixes the issue 👍

src/query.js
```diff
@@ -1024,6 +1024,23 @@ function CreateRole(params) {
...
 }
 
+/**
+ * See the [docs](https://app.fauna.com/documentation/reference/queryapi#write-functions).
+ *
+ * @param {module:query~ExprArg} params
+ *   An object of parameters used to create a new access provider.
+ *     - name: A valid schema name
+ *     - issuer: A unique string
+ *     - jwks_uri: A valid HTTPS URI
```
fauna-ryanharris, line 1034: @nathaniel-may @marrony The JIRA ticket referred to this field as `jwks_url`, but Core has it as `jwks_uri`. Let me know if you want to change it back for any reason 👍

fauna-ryanharris:
> `issuer` cannot be duplicate, feel free to add a test for this if you want, other than that looks good to me.

@marrony I thought that might be the case based on my testing. Thank you for the confirmation -- I'll add the test now 🙏

PR #306, opened 2020-07-02
Fix match and paginate stringify

BrunoQuaresma:


test/values.test.js
```diff
@@ -255,6 +255,37 @@ describe('Values', () => {
...
+  test('pretty print Match', () => {
+    assertPrint(
+      new Query(q.Lambda('_', q.Match(q.Index('idx')))),
+      'Query(Lambda("_", Match(Index("idx"))))'
+    )
+
+    assertPrint(
+      new Query(q.Lambda('_', q.Match(q.Index('idx'), 'str'))),
+      'Query(Lambda("_", Match(Index("idx"), "str")))'
+    )
```
BrunoQuaresma, line 267: @marrony this scenario was the one witch was breaking console for some users.

src/Expr.js
```diff
@@ -128,6 +128,33 @@ var exprToString = function(expr, caller) {
...
+    var terms = expr['terms'] || []
+
+    if (isExpr(terms)) terms = terms.raw
+
+    if (Array.isArray(terms) && terms.length == 0)
+      return 'Match(' + matchStr + ')'
+
+    if (Array.isArray(terms)) {
+      return 'Match(' + matchStr + ', ' + printArray(terms, exprToString) + ')'
+    }
```
BrunoQuaresma, line 142: I added a simple verification to see if the terms is an array before call printArray. Sometimes, terms is not an array when the Match is called using one argument like `Match(Index(), "some")`.

test/values.test.js
```diff
@@ -255,6 +255,37 @@ describe('Values', () => {
...
+
+    assertPrint(
+      new Query(q.Lambda('_', q.Match(q.Index('idx'), 'str', 10))),
+      'Query(Lambda("_", Match(Index("idx"), "str", 10)))'
```
marrony, line 271: Please add a other version of that assertion that pass the arguments to Match as an array

BrunoQuaresma: Will do!

BrunoQuaresma: Done!

test/values.test.js
```diff
@@ -255,6 +255,42 @@ describe('Values', () => {
...
+    )
+
+    assertPrint(
+      new Query(q.Lambda('_', q.Match(q.Index('idx'), ['str', 10]))),
+      'Query(Lambda("_", Match(Index("idx"), "str", 10)))'
```
marrony, line 276: I was expecting `"str", 10` to be renderized as `["str", 10]`, we have a condition on the code to do this

src/Expr.js
marrony: you should add right braces around `printArray` function because it doesn't print it.
```
if (Array.isArray(terms)) {
       return 'Match(' + matchStr + ', [' + printArray(terms, exprToString) + '])'
     }
```

BrunoQuaresma: Usually, I think nested guard clauses are hard to read, what do you think? I also put the `[]` around the `printArray`.

test/values.test.js
```diff
@@ -255,6 +255,42 @@ describe('Values', () => {
...
+    )
+
+    assertPrint(
+      new Query(q.Lambda('_', q.Match(q.Index('idx'), 'str', 10))),
+      'Query(Lambda("_", Match(Index("idx"), ["str", 10])))'
+    )
+
+    assertPrint(
+      new Query(q.Lambda('_', q.Match(q.Index('idx'), ['str', 10]))),
+      'Query(Lambda("_", Match(Index("idx"), ["str", 10])))'
```
marrony, line 276: looking the result, looks good to me, but what you guys think? use brackets always or remove always when rendering the expression? both are equivalente

BrunoQuaresma: I have no strong opinions on that... I would say to use `[]` always to be more explicit? I'll ask on #developer-experience channel to see what other folks think about this.

BrunoQuaresma: @marrony looks like we are going to use `[]` is that ok for you we merge this?

PR #307, opened 2020-07-05
feat: Added an option for automatic retries

tomusiaka2:


nathaniel-may:
Hi @tomusiaka2, thank you for opening this PR. I've raised this internally as a community-driven feature request so it is available for future roadmap planning. There are a few considerations we have to make that may require changes to the fauna service itself so I'm going to close this in favor of developing this feature internally. Thanks for being a part of the Fauna community!

tomusiaka2:
Thanks @nathaniel-may . No problem and thanks for the update. I was also thinking that if you were to provide automatic client retries functionality, you would probably want to include that in all your drivers at the same time and not just JavaScript

PR #308, opened 2020-07-08
FE-694 | Supported versioned @query values

fauna-ryanharris:
### Notes
[Jira Ticket](https://faunadb.atlassian.net/browse/FE-694)

This PR adds support for versions Query/Lambda values.

### How to test
Run the test suite with `npm run test`

### DIFF
--- a/src/Expr.js
+++ b/src/Expr.js
@@ -140,15 +140,13 @@ var exprToString = function(expr, caller) {
 
   if ('object' in expr) return printObject(expr['object'])
 
-  var keys = Object.keys(expr)
-  var fn = keys[0]
-
   // Versioned queries/lambdas will have an api_version field.
   // We want to prevent it from being parsed and displayed as:
   // Query(ApiVersion("3", "X", Var("X")))
-  if (fn === 'api_version') {
-    fn = keys[1]
-  }
+  var keys = Object.keys(expr).filter(
+    expression => expression !== 'api_version'
+  )
+  var fn = keys[0]
 
   // For FQL functions with special formatting concerns, we
   // use the specialCases object above to define their casing.
@@ -161,16 +159,10 @@ var exprToString = function(expr, caller) {
     })
     .join('')
 
-  var args = keys
-    .filter(key => {
-      // We don't want to include api_version values in
-      // our args and subsequent FQL string creation.
-      return key !== 'api_version'
-    })
-    .map(function(k) {
-      var v = expr[k]
-      return exprToString(v, fn)
-    })
+  var args = keys.map(function(k) {
+    var v = expr[k]
+    return exprToString(v, fn)
+  })
 
   var shouldReverseArgs = ['filter', 'map', 'foreach'].some(function(fn) {
     return fn in expr
--- a/src/Expr.js
+++ b/src/Expr.js
@@ -143,6 +143,13 @@ var exprToString = function(expr, caller) {
   var keys = Object.keys(expr)
   var fn = keys[0]
 
+  // Versioned queries/lambdas will have an api_version field.
+  // We want to prevent it from being parsed and displayed as:
+  // Query(ApiVersion("3", "X", Var("X")))
+  if (fn === 'api_version') {
+    fn = keys[1]
+  }
+
   // For FQL functions with special formatting concerns, we
   // use the specialCases object above to define their casing.
   if (fn in specialCases) fn = specialCases[fn]
@@ -154,10 +161,16 @@ var exprToString = function(expr, caller) {
     })
     .join('')
 
-  var args = keys.map(function(k) {
-    var v = expr[k]
-    return exprToString(v, fn)
-  })
+  var args = keys
+    .filter(key => {
+      // We don't want to include api_version values in
+      // our args and subsequent FQL string creation.
+      return key !== 'api_version'
+    })
+    .map(function(k) {
+      var v = expr[k]
+      return exprToString(v, fn)
+    })
 
   var shouldReverseArgs = ['filter', 'map', 'foreach'].some(function(fn) {
     return fn in expr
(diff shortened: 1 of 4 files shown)


test/query.test.js
```diff
@@ -222,6 +222,7 @@ describe('query', () => {
         new values.Query({
           lambda: '_',
           expr: { let: { x: 1, y: 2 }, in: [{ var: 'x' }, { var: 'y' }] },
+          api_version: '3',
```
fauna-ryanharris, line 225: Without the new `api_version` field here, this test was failing

test/query.test.js
```diff
@@ -337,7 +338,7 @@ describe('query', () => {
 
     return client.query(query.Query(lambda)).then(function(body) {
       return client.query(body).then(function(bodyEchoed) {
-        expect(body).toEqual(bodyEchoed)
+        expect(body.value).toEqual(bodyEchoed.value)
```
fauna-ryanharris, line 341: I had to update this assertion since `body` and `bodyEchoed` now both have `api_version` fields, which can mismatch. For now, we only really care about the `value` being equal.

src/values.js
```diff
@@ -343,10 +343,18 @@ Bytes.prototype.toJSON = function() {
 /** FaunaDB query. See the [docs](https://app.fauna.com/documentation/reference/queryapi#special-type).
  *
  * @param {any} value
+ * @param {any} api_version
  * @extends module:values~Value
  * @constructor
  */
 function Query(value) {
+  this.api_version = 'unstable'
```
fauna-ryanharris, line 351: @marrony The ticket mentioned this as a possible value, but should `2.12` be the default?

marrony: users should not be able to set this value, `Query` should be an opaque type

fauna-ryanharris:
@marrony I update the PR based on your notes. Let me know what you think 🙏

src/Expr.js
```diff
@@ -143,6 +143,13 @@ var exprToString = function(expr, caller) {
   var keys = Object.keys(expr)
   var fn = keys[0]
 
+  // Versioned queries/lambdas will have an api_version field.
+  // We want to prevent it from being parsed and displayed as:
+  // Query(ApiVersion("3", "X", Var("X")))
```
marrony, line 148: please, add a test on https://github.com/fauna/faunadb-js/blob/master/test/values.test.js#L146 to make sure this is correct

fauna-ryanharris: @marrony Would this test suffice? https://github.com/fauna/faunadb-js/pull/308/files#diff-c2e271250d0fe43a2f2478970f73543eR2713

marrony: since you mentioned it, that test doesn't make sense in this context because it's not deserializing anything, it would make more sense if you move it to "pretty print" on `values.test.js`

test/query.test.js
```diff
@@ -2702,6 +2703,34 @@ describe('query', () => {
     }
   })
 
+  test('deserialize version lambdas correctly', async () => {
+    try {
+      const rawQuery = {
+        '@query': { api_version: '3', lambda: 'X', expr: { var: 'X' } },
```
marrony, line 2709: you can remove `api_version` attribute here and send it using `client.query()` and make sure it returns `api_version: "2.12"`

test/query.test.js
```diff
@@ -2702,6 +2703,34 @@ describe('query', () => {
...
+      }
+      const query = new values.Query(rawQuery['@query'])
+
+      expect(query.toString()).toBe(`Query(Lambda("X", Var("X")))`)
+    } catch (error) {
+      console.log(error)
+    }
+  })
+
+  test('serialize version lambdas correctly', async () => {
```
marrony, line 2719: the title is misleading because we are not serializing any api version here

test/query.test.js
marrony: I was expecting a separate test specifically for this, similar to `legacy queries/lambdas have default api_version` where you call the `Query()` function and assert the returned value contains the `api_version` equals the latest api version driver is using, this was an accidental coincidence

test/query.test.js
```diff
@@ -2702,6 +2703,35 @@ describe('query', () => {
     }
   })
 
+  test('parse query/lambda version properly', () => {
+    try {
+      const rawQuery = {
+        '@query': { api_version: '3', lambda: 'X', expr: { var: 'X' } },
+      }
+      const newQuery = new values.Query(rawQuery['@query'])
+      const json = newQuery.toJSON()
```
marrony, line 2712: the best place to test `toJSON()` method is on `values.test.js`

test/query.test.js
```diff
@@ -2702,6 +2703,35 @@ describe('query', () => {
...
+    try {
+      const res = await client.query(
+        new values.Query({ lambda: 'X', expr: { var: 'X' } })
+      )
+
+      expect(res).toBeInstanceOf(values.Query)
+      expect(res.toJSON()).toEqual({
+        '@query': { api_version: '2.12', lambda: 'X', expr: { var: 'X' } },
+      })
+    } catch (error) {
```
marrony, line 2730: try/catch on tests only makes sense if you fail the test, as is, `console.log(...)`, the tests will always pass, even though an exception is thrown

fauna-ryanharris:
@marrony I updated the tests based on your latest notes 👍

src/Expr.js
```diff
@@ -143,6 +143,13 @@ var exprToString = function(expr, caller) {
...
+  if (fn === 'api_version') {
```
marrony, line 149: there are no guaranties that `api_version` field will be the first attribute

fauna-ryanharris: @marrony Makes sense. I'll update the PR and ping you when it's ready for review. Thanks for all your help 🙏

fauna-ryanharris: **NOTE:** We are going to enact these changes in a separate PR 👍

PR #309, opened 2020-07-10
FE-702 | docs: add npm pack instructions

fauna-ryanharris:
### Notes
[Jira Ticket](https://faunadb.atlassian.net/browse/FE-702)

This PR adds instructions to the README on how to use `npm pack` in order to create a local tarball and preview unreleased changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -284,7 +284,29 @@ Alpine-based NodeJS image can be provided via `RUNTIME_IMAGE`).
 
 ### Previewing upcoming functionality
 
-If you want to test upcoming features that have yet to be released, you can do the following:
+If you want to preview unreleased features in your project, you can do so by installing this driver using one of the following methods.
+
+#### 1. Using a git URL
+
+Normally, you would install the latest release of this package using `npm install --save faunadb` or `yarn add faunadb`. To access our latest features, you will need to define this dependency [by using a git URL](https://docs.npmjs.com/files/package.json#dependencies).
+
+1. Open your `package.json` file
+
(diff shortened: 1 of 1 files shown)


BrunoQuaresma:
About this, if the functionality is to test only on local development you can also use a path to the project directly on package.json without the need to pack stuff or, you can also point that to git repo.

fauna-ryanharris:
> About this, if the functionality is to test only on local development you can also use a path to the project directly on package.json without the need to pack stuff or, you can also point that to git repo. So I don't know if recommend this flow is the best way to follow 🤔 . Let's wait for more opinions.

True. Users could just point their `package.json` at the GitHub repo's master branch, which would also work. I thought this tarball might be a better way to go since it represents the actual image stored in the NPM registry. However, I'm open to changing this if we want.

README.md
```diff
@@ -282,6 +282,55 @@ Alpine-based NodeJS image can be provided via `RUNTIME_IMAGE`).
...
+
+1. Open your `package.json` file
+
+2. If you have already installed this driver, you should see the following in your list of dependencies. If not, add it.
+
+```
+"faunadb": "^2.14.1"
+```
+
+3. Instead of using a version from the npm registry, we'll want to point our `package.json` to the `master` branch of our GitHub repo. To do that, change the `^2.4.1` to `fauna/faunadb-js#master`.
```
faunaee, line 301: Is it `2.4.1`, or `2.14.1` as above?

PR #310, opened 2020-07-15
FE-705 | Refactor parsing logic to be order agnostic

fauna-ryanharris:
### Notes
[Jira Ticket](https://faunadb.atlassian.net/browse/FE-705)

Currently, the JS driver depends on the order of the arguments it receives as wire protocol, which should not be the case. We should not base the logic on order, so this PR attempts to improve that functionality.

### How to test
Run `npm run test`


test/values.test.js
```diff
@@ -150,6 +150,49 @@ describe('Values', () => {
...
+    var complexQuery = new Query({
+      lambda: '_',
+      expr: { match: { index: 'idx' }, terms: ['str', 10] },
+      api_version: '3',
+    })
+
+    // simpleQuery assertions
+    var x = '{"@query":{"api_version":"3","lambda":"x","expr":{"var":"x"}}}'
+    expect(json.parseJSON(x)).toEqual(simpleQuery)
+    expect(json.parseJSON(x).toString()).toEqual(simpleQuery.toString())
```
fauna-ryanharris, line 169: @marrony Would it be better to have these `.toString()` assertion in the `pretty print` test block below?

marrony: I think the `.toString()` assertions here are not necessary, if you parsed the json and it equals to what you expect, then the `.toString()` would be the same

fauna-ryanharris: @marrony That's what I was thinking as well -- just trying to cover each case as well as I can. I'll remove these 👍

src/Expr.js
```diff
@@ -169,35 +236,99 @@ var exprToString = function(expr, caller) {
...
+    )
+  }
+
+  if ('call' in expr) {
+    if (Array.isArray(expr.arguments.raw)) {
+      return (
+        'Call(' +
+        exprToString(expr['call']) +
+        ', ' +
+        printArray(expr['arguments']['raw'], exprToString) +
```
marrony, line 297: instead of introspect `raw` field here, you should call the function `exprToString` recursively and it should resolve the value, in the beginning of the function it already does this check

test/values.test.js
```diff
@@ -150,6 +150,86 @@ describe('Values', () => {
...
+    var b =
+      '{"@query":{"expr":{"match":{"index":"idx"},"terms":["str",10]},"api_version":"3","lambda":"_"}}'
+    expect(json.parseJSON(b)).toEqual(complexQuery)
+
+    var c =
+      '{"@query":{"api_version":"3","expr":{"match":{"index":"idx"},"terms":["str",10]},"lambda":"_"}}'
+    expect(json.parseJSON(c)).toEqual(complexQuery)
+  })
+
+  test('parse functions regardless of order', () => {
```
marrony, line 190: this test is not necessary because the driver doesn't parse the expression, you should focus on the pretty print tests to ensure the queries values render the correct code

fauna-ryanharris: @marrony I was going to ask you about that 😅 Should I just remove this whole `test` block?

marrony: fell free to remove

src/Expr.js
```diff
@@ -169,35 +236,109 @@ var exprToString = function(expr, caller) {
...
+      ', ' +
+      exprToString(expr['else']) +
+      ')'
+    )
+  }
+
+  if ('call' in expr) {
+    var argumentsExpr = null
+
+    if (isExpr(expr['arguments'])) {
```
marrony, line 294: this whole logic is not necessary, just calling `argumentsExpr = exprToString(expr['arguments'])` should be enough, all the logic added here was already handled in the beginning of the function

src/Expr.js
```diff
@@ -169,35 +236,109 @@ var exprToString = function(expr, caller) {
...
+        argumentsExpr = '[' + printArray(expr['arguments'], exprToString) + ']'
+      } else if (typeof expr['arguments'] === 'object') {
+        argumentsExpr = printObject(expr['arguments'])
+      } else {
+        argumentsExpr = exprToString(expr['arguments'])
+      }
+    }
+
+    var callExpr = ''
+    if (isExpr(expr['call'])) {
```
marrony, line 307: no need to check if `expr['call']` is an expression, the recursion will already do this

src/Expr.js
```diff
@@ -169,35 +236,109 @@ var exprToString = function(expr, caller) {
...
+      callExpr = exprToString(expr['call'])
+    } else {
+      callExpr = exprToString(expr['call'])
+    }
+
+    return 'Call(' + callExpr + ', ' + argumentsExpr + ')'
```
marrony, line 313: this can be simplified by
```javascript
return 'Call(' + exprToString(expr['call']) + ', ' + exprToString(expr['arguments']) + ')'
```

src/Expr.js
fauna-ryanharris: Ah, true! Thanks. Will update.

src/Expr.js
fauna-ryanharris: I added this check because one of the pre-existing unit tests was getting to this point where the `arguments` and `call` values were both expressions 🤔

src/Expr.js
```diff
@@ -169,35 +228,109 @@ var exprToString = function(expr, caller) {
...
+    if (isExpr(expr['arguments'])) {
+      argumentsExpr = exprToString(expr['arguments'])
+    } else {
+      if (Array.isArray(expr['arguments'])) {
+        argumentsExpr = '[' + printArray(expr['arguments'], exprToString) + ']'
+      } else if (typeof expr['arguments'] === 'object') {
+        argumentsExpr = printObject(expr['arguments'])
+      } else {
+        argumentsExpr = exprToString(expr['arguments'])
+      }
```
marrony, line 295: not necessary too

src/Expr.js
```diff
@@ -169,35 +228,114 @@ var exprToString = function(expr, caller) {
...
+  }
+
+  if ('call' in expr) {
+    var argumentExpr = exprToString(expr['arguments'])
+
+    // Print when arguments are an object
+    if (
+      typeof expr['arguments'] === 'object' &&
+      !Array.isArray(expr['arguments']) &&
+      !expr['arguments']['raw']
```
marrony, line 300: that condition should never happen, you will never receive a vanilla object here, objects in our protocol are wrapped in `{"object": {...}}` function, and again, it will be handled in the beginning of the function

test/values.test.js
```diff
@@ -501,12 +538,176 @@ describe('Values', () => {
...
+        call: {
+          function: 'double',
+        },
+      }),
+      `Query(Call(Function("double"), [1, 2, 3]))`
+    )
+
+    assertPrint(
+      new Query({
+        arguments: {
```
marrony, line 701: that argument is not correct, objects are wrapped in the protocol, the correct would be
```
new Query({
  arguments: {
    object: {
      one: 1,
      two: 2,
      three: 3
    }
  },
  call: {
    function: 'double'
  }
})
```

src/Expr.js
```diff
@@ -169,35 +228,114 @@ var exprToString = function(expr, caller) {
...
+
+    return 'Call(' + exprToString(expr['call']) + ', ' + argumentExpr + ')'
+  }
+
   // Versioned queries/lambdas will have an api_version field.
-  // We want to prevent it from being parsed and displayed as:
-  // Query(ApiVersion("3", "X", Var("X")))
+  // We want to prevent it from being parsed and displayed.
   var keys = Object.keys(expr).filter(
     expression => expression !== 'api_version'
```
marrony, line 311: given lambdas are already handled before this point, that filter is not necessary

src/Expr.js
```diff
@@ -169,35 +228,114 @@ var exprToString = function(expr, caller) {
...
     var v = expr[k]
     return exprToString(v, fn)
   })
 
-  var shouldReverseArgs = ['filter', 'map', 'foreach'].some(function(fn) {
-    return fn in expr
-  })
+  // Return string without args to avoid showing null args
+  if (args.length === 1 && args[0] === 'null') {
+    return fn + '()'
```
marrony, line 335: this will cause incorrect behaviors, not all functions with 1-arity that receives `null` should be rendered as `fn()`, for example, `IsNumber(null)` as is will be rendered as `IsNumber()`

src/Expr.js
```diff
@@ -169,35 +228,114 @@ var exprToString = function(expr, caller) {
...
   )
+
   var fn = keys[0]
 
-  // For FQL functions with special formatting concerns, we
-  // use the specialCases object above to define their casing.
-  if (fn in specialCases) fn = specialCases[fn]
+  // Handle functions with an arity greater than 1
+  if (keys.length > 1) {
+    fn = getFunctionFromKeys(keys)
```
marrony, line 318: I don't think I understood the purpose of function `getFunctionFromKeys`, all the functions with arity > 1 will be handled before this point to avoid order problems

test/values.test.js
fauna-ryanharris: Ah, okay. Thank you. I was missing the `object` 🤦

src/Expr.js
fauna-ryanharris: Can you clarify this a bit more? My understanding was that whenever a function has one argument which is `null`, we don't want to render that.

src/Expr.js
```diff
@@ -169,35 +218,92 @@ var exprToString = function(expr, caller) {
...
+    return (
+      'Call(' +
+      exprToString(expr['call']) +
+      ', ' +
+      exprToString(expr['arguments']) +
+      ')'
+    )
+  }
+
+  if ('databases' in expr) {
```
fauna-ryanharris, line 293: @marrony I added this to handle the case where tests were failing due to `Databases(null)` being rendered. I'm assuming we'd also want to include `Collections()`, `Documents()`, etc. as well, right?

marrony: yes, that's correct

src/Expr.js
```diff
@@ -74,60 +74,109 @@ var specialCases = {
...
+}
 
-  if (type === 'undefined') {
-    return 'undefined'
+/**
+ *
+ * @param {*} expr A FQL expression
+ * @returns {Boolean} Indicates if arguments should be reversed
+ */
+function shouldReverseArgs(expr) {
```
marrony, line 143: also, once `exprToString` is finished, that function won't necessary, `filter` and `foreach` is already handled there, `map` should be too

src/Expr.js
marrony: only 0-arity functions should drop null, like `Now()`. `IsNumber(x)` is 1-arity function and whatever argument was passed in should be preserved. You need to look at the wire to protocol to understand better, given `{}` represents a function call and the fields represents the function/arguments name, `Now()` function becomes `{"now": null}` in the wire, the null value is there because something must be in the value. On the other hand, `IsNumber(x)` is a 1-arity function, and it becomes `{"is_number": x}` in the wire

src/Expr.js
marrony: bear in mind that `Databases()` and its siblings could receive an argument, the scope, so you should check if `expr[databases]` is not null and render it too like `Databases(Database('child-db'))`

test/values.test.js
```diff
@@ -501,12 +538,178 @@ describe('Values', () => {
...
+        collection: [1, 2, 3],
+        api_version: '3',
+        filter: {
+          expr: {
+            equals: [0, { modulo: [{ var: 'i' }, 2] }],
+          },
+          lambda: 'i',
+        },
+      }),
+      'Query(Filter([1, 2, 3], Lambda("i", Equals(0, Modulo(Var("i"), 2)))))'
```
marrony, line 597: although this is kinda acceptable for testing purposes, you will never receive a `@query` type with a `Filter` expression. Queries types only wrap Lambdas

test/values.test.js
```diff
@@ -501,12 +538,178 @@ describe('Values', () => {
...
+        lambda: '_',
+      }),
+      'Query(Lambda("_", Match(Index("idx"), 10)))'
+    )
+
+    // Filter an array
+    assertPrint(
+      new Query({
+        collection: [1, 2, 3],
+        api_version: '3',
```
marrony, line 589: Filter expressions doesn't have `api_version` argument.

src/Expr.js
fauna-ryanharris: To make sure I understand, was this originally here to force the order of arguments for `filter`, `foreach` and `map` but now we don't need this function since our `if` statements are now handling constructing the queries properly?

test/values.test.js
fauna-ryanharris: @marrony I noticed this when testing queries in the webshell that no `@query` was coming back for certain things, but my assertions were failing for some reason without `Query()`. I'm happy to go back and update these if you'd like 👍

test/values.test.js
fauna-ryanharris: Good call. I'll remove this 👍

src/Expr.js
marrony: That’s correct

test/values.test.js
marrony: I don’t think it’s necessary to change, it was just a heads up to make sure you add/keep the lambdas tests

fauna-ryanharris: In the future, would the best way to do it be using `new Expr()`?

marrony: `Query` extends from `Expr`

test/values.test.js
```diff
@@ -501,12 +538,228 @@ describe('Values', () => {
...
+        collection: { paginate: { documents: { collection: 'orders' } } },
+        foreach: {
+          lambda: 'doc',
+          expr: {
+            update: { var: 'doc' },
+            params: { object: { data: { object: { createdBy: 'Ryan' } } } },
+          },
+        },
+      }),
+      `Query(Foreach(Paginate(Documents(Collection("orders"))), Lambda("doc", Update(Var("doc"), {data: {createdBy: "Ryan"}}))))`
```
fauna-ryanharris, line 664: @marrony I hit some of your notes re: Databases, Collections and Documents, however, this test is now failing. For some reason, the Documents() never displays the Collection("orders") inside it.

I've tracked the issue to the `if ('paginate' in expr)` block of Expr.js, but am not sure why it is disregarding my current logic (which I think should work) 🤔

src/Expr.js
fauna-ryanharris: Some tests are now failing if I remove this function call.

src/Expr.js
```diff
@@ -169,35 +230,108 @@ var exprToString = function(expr, caller) {
...
+  if ('collections' in expr) {
+    if (!expr['database']) return 'Collections()'
+
+    return 'Collections(' + exprToString(expr['collections']) + ')'
+  }
+
+  if ('documents' in expr) {
+    console.log('in documents block')
+    console.log(expr)
+    if (!expr['collection']) return 'Documents()'
```
marrony, line 320: No need to test this, `collection` will never appear along with `documents` expression

src/Expr.js
```diff
@@ -169,35 +230,108 @@ var exprToString = function(expr, caller) {
...
+      'Call(' +
+      exprToString(expr['call']) +
+      ', ' +
+      exprToString(expr['arguments']) +
+      ')'
+    )
+  }
+
+  if ('databases' in expr) {
+    if (!expr['database']) return 'Databases()'
```
marrony, line 306: `expr['databases']`

src/Expr.js
```diff
@@ -169,35 +230,108 @@ var exprToString = function(expr, caller) {
...
+
+    return 'Databases(' + exprToString(expr['databases']) + ')'
+  }
+
+  if ('collections' in expr) {
+    if (!expr['database']) return 'Collections()'
```
marrony, line 312: `expr['collections']`

src/Expr.js
marrony: you should add `map` to the if wagon

test/values.test.js
marrony: but it could probably be a good idea to start using `Expr` like those tests https://github.com/fauna/faunadb-js/pull/310/files#diff-04f5179b95b9dabc1a5d3bfa30234fffR716

src/Expr.js
```diff
@@ -169,35 +225,115 @@ var exprToString = function(expr, caller) {
...
+  }
+
+  if ('collections' in expr) {
+    if (!expr['collections']) return 'Collections()'
+
+    return 'Collections(' + exprToString(expr['collections']) + ')'
+  }
+
+  if ('documents' in expr) {
+    if (!expr['documents']) return 'Documents()'
```
marrony, line 313: `Documents` functions always receive an argument, there is not such overload with 0-arity

src/Expr.js
```diff
@@ -169,35 +225,115 @@ var exprToString = function(expr, caller) {
...
     var v = expr[k]
     return exprToString(v, fn)
   })
 
-  var shouldReverseArgs = ['filter', 'map', 'foreach'].some(function(fn) {
-    return fn in expr
-  })
-
-  if (shouldReverseArgs) {
+  if (shouldReverseArgs(expr)) {
```
marrony, line 336: now that you handled all the edge case, that function call and the function itself is not necessary

test/values.test.js
```diff
@@ -501,12 +538,228 @@ describe('Values', () => {
...
+        },
+      }),
+      'Query(Documents(Collection("lettuce")))'
+    )
+
+    assertPrint(
+      new Query({
+        documents: null,
+      }),
+      'Query(Documents())'
```
marrony, line 762: that assertion is not correct, if `documents` field is null, then it should render `Documents(null)`

fauna-ryanharris: @marrony Ah, okay. I was wondering why that was failing 😅

src/Expr.js
fauna-ryanharris: Done 👍

src/Expr.js
```diff
@@ -169,38 +214,113 @@ var exprToString = function(expr, caller) {
...
+      ')'
+    )
+  }
 
-  // For FQL functions with special formatting concerns, we
-  // use the specialCases object above to define their casing.
-  if (fn in specialCases) fn = specialCases[fn]
+  if ('containsstr' in expr) {
+    return (
+      'ContainsStr(' +
```
marrony, line 229: are you planning to add other string functions here? `containsstr` is already handled by `specialCases` map, if you plan to add those functions here, you can remove from there

fauna-ryanharris: @marrony I guess we could remove them from the special cases map and add all the `Contains` functions here since we need to make sure the argument order is constructed properly 👍 

What do you think?

ecwyne:
@fauna-ryanharris I just wanted to double check on something.

Currently I can do the following:
```js
q.Let({
    one: q.Add(0, 1),
    two: q.Add(q.Var('one'), 1),
    three: q.Add(q.Var('two'), 1),
}, ...)
```

Will this still be possible with these changes?

fauna-ryanharris:
> @fauna-ryanharris I just wanted to double check on something.
> 
> Currently I can do the following:
> 
> ```js
> q.Let({
>     one: q.Add(0, 1),
>     two: q.Add(q.Var('one'), 1),
>     three: q.Add(q.Var('two'), 1),
> }, ...)
> ```
> 
> Will this still be possible with these changes?

@ecwyne I believe so as I did not change any of the `let` functionality. @marrony Is there any reason Eric wouldn't be able to continue doing this?

ecwyne:
To clarify, I only ask because the order of the properties on the `Let` object matters

marrony:
@ecwyne this change has nothing to do with fql syntax per se, this change is only related to the fauna shell and webshell when the driver render a lambda type, it should produce an executable code so the user can copy and paste, or edit the lambda predicates on abac ui. @fauna-ryanharris Eric brought an interesting point, you need to make sure the let doesnt change the order of the variable declared.

BrunoQuaresma:
@ryancharris I think add a test verifying the order on let params can be enough to ensure that.

test/query.test.js
```diff
@@ -225,6 +225,17 @@ describe('query', () => {
           api_version: '3',
         })
       ),
+      assertQuery(
+        query.Let(
```
fauna-ryanharris, line 229: @marrony @BrunoQuaresma Changing the order of the fields in this object causes the test to fail. However, I don't think we've necessarily introduced this as a bug 🤔 Thoughts?

BrunoQuaresma: This test looks good :D

test/values.test.js
```diff
@@ -501,12 +538,265 @@ describe('Values', () => {
...
+    assertPrint(
+      new Query({
+        let: {
+          three: { add: [{ var: 'two' }, 1] },
+          two: { add: [{ var: 'one' }, 1] },
+          one: { add: [0, 1] },
+        },
+        in: { var: 'three' },
+      }),
+      'Query(Let({three: Add(Var("two"), 1), two: Add(Var("one"), 1), one: Add(0, 1)}, Var("three")))'
```
marrony, line 799: I would add a test with arrays to make sure we are printing it correctly, something like: `Query(Let([{x: 10}, {y: 20}], Add(Var('x'), Var('y'))))`

fauna-ryanharris: @marrony This query doesn't work in the Webshell. From my reading of the FQL cheatsheet, it seems the second arg to `Let` needs to be a `Var` not an `Add`.

marrony: the second argument of `Let` is an expression, so it basically can be anything

fauna-ryanharris: @marrony Got it. I expected it to work, but I am getting this error:
```
Error: [
  {
    "position": [
      "query"
    ],
    "code": "invalid expression",
    "description": "No form/function found, or invalid argument keys: { let, in }."
  }
]
```

marrony: this worked for me
```javascript
Query(Lambda(x => Let([{y: x}], Add(1, Var("y")))))
```

fauna-ryanharris:
@marrony Double checking to see if you have any other notes before I merge this in 👍

marrony:
well, given we are now rendering lambdas so that users can persist and later edit it on the UI, ideally we should add tests with all functions in the API

fauna-ryanharris:
> well, given we are now rendering lambdas so that users can persist and later edit it on the UI, ideally we should add tests with all functions in the API

cc: @danieltodonnell @nathaniel-may Should we include this in within the scope of this PR or spin the work off into its own ticket?

danieltodonnell:
@ryancharris  My vote is to spin it into it's own card.  AS it is, THIS card was already overreach from the APIv3 work.  Let's let Product prioritize it.

marrony:
sounds good to me, this PR let the current code on a par with the previous one, so any "bugs" as of now will likely exist anyway

fauna-ryanharris:
Thanks for all your help with this PR, @marrony . Much appreciated 🙏 I will queue up another ticket for adding tests to cover the Lambdas 👍

fauna-ryanharris:
> Thanks for all your help with this PR, @marrony . Much appreciated 🙏 I will queue up another ticket for adding tests to cover the Lambdas 👍

Here is the follow-up ticket: https://faunadb.atlassian.net/browse/FE-728

PR #312, opened 2020-07-21
FE-613 | Move release pipeline to GH Actions

fauna-ryanharris:
### Notes
[Jira Ticket](https://faunadb.atlassian.net/browse/FE-613)

**To Do:**

- [x] Configure `semantic-release` with proper npm and GitHub tokens

- [x] Save tokens as env vars for access during workflow runs


.github/workflows/create-release.yml
```diff
@@ -0,0 +1,22 @@
+name: Create GitHub release
+
+on:
```
fauna-ryanharris, line 3: We'll be changing this trigger to use `repository_dispatch` so we can trigger it with HTTP requests sent from Blue 🐦

.github/workflows/create-release.yml
```diff
@@ -0,0 +1,22 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Create release
+        uses: actions/create-release@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }} # This token is provided by Actions, you do not need to create your own
+        with:
+          tag_name: ${{ github.ref }}
```
fauna-ryanharris, line 19: Since the trigger won't be a new tag, I'll need to figure out how to parse the tag name here and below

fauna-ryanharris: From what I've read in the docs thus far, I believe we can send the version tag in the body of the HTTP request and parse its value here to use for creating out GitHub tag 👍

.github/workflows/semantic-release.yml
```diff
@@ -0,0 +1,27 @@
+name: Semantic release
+
+on:
+  repository_dispatch:
+    types: [semantic-release]
```
fauna-ryanharris, line 5: This is triggered from our Blue bot. The PR for that work can be found here: https://github.com/fauna/blue/pull/1

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "faunadb",
-  "version": "2.14.2",
+  "version": "0.0.0-development",
```
fauna-ryanharris, line 3: `semantic-release` changes the version in our `package.json` but handles updating the version when publishing based on the most recent commit tag/our current version on NPM

fauna-ryanharris: I'm going to read the docs a bit more and see if we can keep our current version or if that **must** change. Don't want to cause any unforeseen publishing issues 😓

fauna-rodrigo: @ryancharris When we add the same thing for console, we were able to keep the version in package.json an let the tool handle it. we did not have issues so far, maybe you can try to let the version field as it is on package.json, and generate a version locally using the --dry-run flag and see if it picks the correct number.

fauna-ryanharris: @fauna-rodrigo Good call. I'll try that 👍 My understanding was that we could keep the version and `semantic-release` would be smart enough to bump the number appropriately based on the commit tag of the most recent commit on master.

fauna-ryanharris:
@fauna-rodrigo I added a `release` config to the package.json, so it'll only do a dry run for now. I think the easiest way to test this will be to merge the PR and then send requests since I need the token values exposed in the GH Action runtime 👍 

What do you think?

PR #314, opened 2020-07-23
Flow types

michalsanger:
This PR adds Flow types to improve DX in projects typed with Flow.

I also added commands into `package.json` to validate type definitions and fixed one bug in TS types.

### How to test
`yarn typecheck`

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -1,0 +1,12 @@
+[ignore]
+<PROJECT_ROOT>/node_modules/
+
+[include]
+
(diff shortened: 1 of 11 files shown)


fauna-ryanharris:
@michalsanger Thank you for your work on this and putting up a PR 🙏 We currently do not have plans to support Flow typing, so I am going to close this out, but we'll have it on record in case we choose to revisit this in the future.

Thanks again! 🚀

PR #319, opened 2020-08-10
chore: update release script

fauna-ryanharris:
### Notes
Last week, we tested this workflow with the `dryRun` flag and `semantic-release` was picking up v2.10.0 as our most recent tag and using that to decide what the new version number would be. This is strange since `git tag` and GitHub both show that our most recent tag is `2.14.2`.

In reading through the documentation, this is because semantic-release uses the tag format `vX.Y.Z` by default whereas our tags use `X.Y.Z` format.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "browserify-min": "browserify index.js --standalone faunadb | terser -c -m --keep-fnames --keep-classnames -o dist/faunadb-min.js",
     "prettify": "prettier --write \"{src,test}/**/*.{js,ts}\"",
     "test": "jest --env=node --verbose=false ./test",
-    "semantic-release": "semantic-release"
+    "semantic-release": "semantic-release --tag-format ${version}"
   },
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -26,7 +26,7 @@
     "browserify-min": "browserify index.js --standalone faunadb | terser -c -m --keep-fnames --keep-classnames -o dist/faunadb-min.js",
     "prettify": "prettier --write \"{src,test}/**/*.{js,ts}\"",
     "test": "jest --env=node --verbose=false ./test",
-    "semantic-release": "semantic-release"
+    "semantic-release": "semantic-release --tag-format ${version}"
```
fauna-ryanharris, line 29: @fauna-rodrigo See these docs: https://semantic-release.gitbook.io/semantic-release/usage/configuration#tagformat

I was originally trying to add the flag in the `"release"` field like our other options, however, the docs don't show how to reference the tag form properly in the JSON 🤷

PR #328, opened 2020-08-25
FE-749 | Revert "chore: refactor parsing logic to be order agnostic (#310)"

fauna-ryanharris:
This reverts commit a65140e0ec8672007f110b8ad6b9d5728e988750.

### Notes
[Jira Ticket](https://faunadb.atlassian.net/browse/FE-749)

In this PR, we added additional logic to make sure the JS driver handled the wire protocol properly if function arguments were out of order. We are now enforcing the order in Core, so no longer need this additional logic.

### How to test

Run `npm run test`


fauna-ryanharris:
@marrony @nathaniel-may I'm reverting these changes to reduce the chance of bugs. Should I keep the tests?

marrony:
there are some tests that won't make sense, mainly the ones that assert the output regardless the order of fields

fauna-ryanharris:
I'm going to close this in favor of a different branch/PR. Instead of reverting the entire commit, I'm going to pull things out manually because I want to keep some function abstractions and new unit tests 👍

PR #329, opened 2020-08-25
FE-749 | Remove unnecessary logic from JS Driver refactor

fauna-ryanharris:
### Notes
[Jira Ticket](https://faunadb.atlassian.net/browse/FE-749)

In #310 , we added additional logic to make sure the JS driver handled the wire protocol properly if function arguments were out of order. We are now enforcing the order in Core, so no longer need this additional logic.

### How to test

Run `npm run test`


fauna-ryanharris:
cc: @nathaniel-may @marrony

PR #330, opened 2020-08-26
FE-750 | Remove special cases from array

fauna-ryanharris:
### Notes
[Jira Ticket](https://faunadb.atlassian.net/browse/FE-750)

In [this PR](https://faunadb.atlassian.net/browse/FE-750), we updated our function names in Core to all be snake_cased. This means we can remove a number of "special cases" from our lookup which handles improperly formatted function names that come over the wire.

### How to test
Run `npm run test`

### DIFF
--- a/src/Expr.js
+++ b/src/Expr.js
@@ -52,25 +52,11 @@ var varArgsFunctions = [
 // Core update: https://faunadb.atlassian.net/browse/ENG-2110
 
 var specialCases = {
-  containsstrregex: 'ContainsStrRegex',
-  endswith: 'EndsWith',
-  findstr: 'FindStr',
-  findstrregex: 'FindStrRegex',
   gt: 'GT',
   gte: 'GTE',
   is_nonempty: 'is_non_empty',
-  lowercase: 'LowerCase',
   lt: 'LT',
   lte: 'LTE',
-  ltrim: 'LTrim',
-  rtrim: 'RTrim',
-  regexescape: 'RegexEscape',
-  replacestr: 'ReplaceStr',
-  replacestrregex: 'ReplaceStrRegex',
-  startswith: 'StartsWith',
-  substring: 'SubString',
-  titlecase: 'TitleCase',
-  uppercase: 'UpperCase',
 }
 
 /**
--- a/src/Expr.js
+++ b/src/Expr.js
@@ -249,6 +249,70 @@ var exprToString = function(expr, caller) {
     )
   }
 
+  if ('collections' in expr) {
+    if (expr['collections'] === null) {
+      return 'Collections()'
+    }
+  }
+
+  if ('databases' in expr) {
+    if (expr['databases'] === null) {
+      return 'Databases()'
+    }
+  }
+
+  if ('select' in expr) {
+    return (
+      'Select(' +
+      exprToString(expr['select']) +
+      ', ' +
+      exprToString(expr['from']) +
+      ')'
+    )
+  }
+
+  if ('if' in expr) {
+    return (
+      'If(' +
+      exprToString(expr['if']) +
+      ', ' +
+      exprToString(expr['then']) +
+      ', ' +
+      exprToString(expr['else']) +
+      ')'
+    )
+  }
+
+  if ('contains_path' in expr) {
+    return (
+      'ContainsPath(' +
+      exprToString(expr['contains_path']) +
+      ', ' +
+      exprToString(expr['in']) +
+      ')'
+    )
+  }
+
+  if ('contains_field' in expr) {
+    return (
+      'ContainsField(' +
+      exprToString(expr['contains_field']) +
+      ', ' +
+      exprToString(expr['in']) +
+      ')'
+    )
+  }
+
+  if ('contains_value' in expr) {
+    return (
+      'ContainsValue(' +
+      exprToString(expr['contains_value']) +
+      ', ' +
+      exprToString(expr['in']) +
+      ')'
+    )
+  }
+
   var keys = Object.keys(expr)
   var fn = keys[0]
   fn = convertToCamelCase(fn)
--- a/src/Expr.js
+++ b/src/Expr.js
@@ -261,6 +261,24 @@ var exprToString = function(expr, caller) {
     }
   }
 
+  if ('functions' in expr) {
+    if (expr['functions'] === null) {
+      return 'Functions()'
+    }
+  }
+
+  if ('indexes' in expr) {
+    if (expr['indexes'] === null) {
+      return 'Indexes()'
+    }
+  }
+
+  if ('roles' in expr) {
+    if (expr['roles'] === null) {
+      return 'Roles()'
+    }
+  }
+
   if ('select' in expr) {
     return (
       'Select(' +
--- a/test/query.test.js
+++ b/test/query.test.js
@@ -419,10 +419,9 @@ describe('query', () => {
   test('take', () => {
     var p1 = assertQuery(query.Take(1, [1, 2]), [1])
     var p2 = assertQuery(query.Take(3, [1, 2]), [1, 2])
-    // var p3 = assertQuery(query.Take(-1, [1, 2]), [])
+    var p3 = assertQuery(query.Take(-1, [1, 2]), [2])
 
-    // return Promise.all([p1, p2, p3])
-    return Promise.all([p1, p2])
+    return Promise.all([p1, p2, p3])
   })
 
   test('drop', () => {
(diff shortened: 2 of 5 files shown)


marrony:
you should update the query.js to use the new symbols
https://github.com/fauna/faunadb-js/blob/master/src/query.js#L1296

fauna-ryanharris:
> you should update the query.js to use the new symbols
> https://github.com/fauna/faunadb-js/blob/master/src/query.js#L1296

Totally! Thanks for catching that 🙏

fauna-ryanharris:
@marrony When I change the values in `query.js`, the unit tests fail. Upon inspecting the error, it seems Core isn't recognizing the snake-cased function names. I've tried testing with our latest stable and nightly builds -- both failed locally.

Is this something that needs to be updated on the Core side?

marrony:
this was not officially released yet, so you need to change the api version on driver to use `UNSTABLE`

test/query.test.js
```diff
@@ -419,9 +419,10 @@ describe('query', () => {
   test('take', () => {
     var p1 = assertQuery(query.Take(1, [1, 2]), [1])
     var p2 = assertQuery(query.Take(3, [1, 2]), [1, 2])
-    var p3 = assertQuery(query.Take(-1, [1, 2]), [])
+    // var p3 = assertQuery(query.Take(-1, [1, 2]), [])
```
fauna-ryanharris, line 422: @marrony Have we made changes to the Take function? This third test no longer passes, however, this seems to be correct based on how Take currently works in the webshell.

test/query.test.js
```diff
@@ -1568,13 +1569,13 @@ describe('query', () => {
...
   })
 
-  test('contains', () => {
-    var obj = { a: { b: 1 } }
-    var p1 = assertQuery(query.Contains(['a', 'b'], obj), true)
-    var p2 = assertQuery(query.Contains('a', obj), true)
-    var p3 = assertQuery(query.Contains(['a', 'c'], obj), false)
-    return Promise.all([p1, p2, p3])
-  })
+  // test('contains', () => {
```
fauna-ryanharris, line 1572: @marrony This no longer passes either. I'm assuming it's because we're removing it in the next API version?

test/query.test.js
```diff
@@ -2324,19 +2325,19 @@ describe('query', () => {
...
-        query.Bytes(new Uint8Array([0, 0, 0, 0])),
-        new Bytes('AAAAAA==')
-      ),
-      assertQuery(query.Bytes(new ArrayBuffer(4)), new Bytes('AAAAAA==')),
-      assertQuery(new Bytes('AQIDBA=='), new Bytes('AQIDBA==')),
-      assertQuery(new Uint8Array([0, 0, 0, 0]), new Bytes('AAAAAA==')),
-      assertQuery(new ArrayBuffer(4), new Bytes('AAAAAA==')),
-    ])
-  })
+  // test('bytes', () => {
```
fauna-ryanharris, line 2328: @marrony Same question here. I'm seeing the following error: `Invalid string. Length must be a multiple of 4`

test/query.test.js
marrony: `Take` will accept negative numbers, this means it will take from the end of the array

test/query.test.js
marrony: you can just remove this test, `Contains` function will be removed on api v4

test/query.test.js
marrony: this is a base64 error on the driver side, you might need to experiment another one that support base64 url safe

fauna-ryanharris: @marrony I don't see any documentation on our site for `Bytes`. Has this been deprecated? If not, what is the expected behavior?

marrony (approved):

ryancharris:
@marrony All of these tests are passing locally, however, CI is failing. Any idea why that might be? I'm assuming it has to do with the Docker image 🤷

BrunoQuaresma:
@ryancharris probably your docker image version is different from the used in the circle CI. Related to the error, I think the test should fails if the "stringify" is returning `null` instead of just an "empty" thing.

```
 Expected: "Query(Filter(Paginate(Databases(null)), Lambda(\"i\", Containsstr(Select(\"name\", Get(Var(\"i\"))), \"-\"))))"
 Received: "Query(Filter(Paginate(Databases()), Lambda(\"i\", ContainsStr(Select(\"name\", Get(Var(\"i\"))), \"-\"))))"
```

The expected is `Databases(null)` but definitely should be `Databases()`

ryancharris:
> @ryancharris probably your docker image version is different from the used in the circle CI. Related to the error, I think the test should fails if the "stringify" is returning `null` instead of just an "empty" thing.
> 
> ```
>  Expected: "Query(Filter(Paginate(Databases(null)), Lambda(\"i\", Containsstr(Select(\"name\", Get(Var(\"i\"))), \"-\"))))"
>  Received: "Query(Filter(Paginate(Databases()), Lambda(\"i\", ContainsStr(Select(\"name\", Get(Var(\"i\"))), \"-\"))))"
> ```
> 
> The expected is `Databases(null)` but definitely should be `Databases()`

@BrunoQuaresma I saw the discrepancy re: `Databases()` and `Databases(null)` 👍

I thought I was using the same Docker image locally, but it seems that's not the case. I'll investigate a bit further.

ryancharris:
@BrunoQuaresma According to this line in the Circle config, it seems I am using the correct `enterprise/nightly:latest` images 🤔 

```
- image: gcr.io/faunadb-cloud/faunadb/enterprise/<<parameters.version>>:latest
```

BrunoQuaresma:
> gcr.io/faunadb-cloud/faunadb/enterprise/<<parameters.version>>:latest

Can you run and see what happens:
```
docker pull gcr.io/faunadb-cloud/faunadb/enterprise/enterprise/nightly:latest
yarn test # or the test command
```

BrunoQuaresma (dismissed):
Code looks good however, I think we need to think in a better way to not merge on master a driver version which is broken on stable versions 🤔 or maybe we cant think on use a different image for tests. @fauna-ryanharris  @nathaniel-may what do you think about this?

ryancharris:
> Code looks good however, I think we need to think in a better way to not merge on master a driver version which is broken on stable versions 🤔 or maybe we cant think on use a different image for tests. @fauna-ryanharris @nathaniel-may what do you think about this?

Agreed. I needed to update the API header to get the tests to pass, but then we're leaving `master` with the incorrect header. I'm open to suggestions for how best to manage this.

src/Expr.js
```diff
@@ -263,6 +249,70 @@ var exprToString = function(expr, caller) {
...
+  if ('contains_value' in expr) {
+    return (
+      'ContainsValue(' +
+      exprToString(expr['contains_value']) +
+      ', ' +
+      exprToString(expr['in']) +
+      ')'
+    )
+  }
+
```
marrony, line 315: no need to add those special cases for those function it is already sneak cased, and the driver logic will convert it to camel case

src/Expr.js
```diff
@@ -263,6 +249,70 @@ var exprToString = function(expr, caller) {
...
+    }
+  }
+
+  if ('databases' in expr) {
+    if (expr['databases'] === null) {
+      return 'Databases()'
+    }
+  }
+
+  if ('select' in expr) {
```
marrony, line 264: doesn't need special case too, also, if it's needed for some reason, the function is missing the `default` argument

src/Expr.js
```diff
@@ -263,6 +249,70 @@ var exprToString = function(expr, caller) {
...
+    if (expr['collections'] === null) {
+      return 'Collections()'
+    }
+  }
+
+  if ('databases' in expr) {
+    if (expr['databases'] === null) {
+      return 'Databases()'
+    }
+  }
```
marrony, line 262: we are going to need special case to remove `null` from all schema types, not only databases and collections

nathaniel-may:
One option would be to use a new branch for the API version dev work then merge that into master once the whole version is complete.

src/Expr.js
ryancharris: @marrony I had originally removed these, however, when I run my unit tests locally with the `nightly:latest` image, these are failing 🤔 Any idea why that might be?

src/Expr.js
ryancharris: @marrony Same here -- this is failing:
image

marrony: yeah, but your test `pretty print despite argument order` is not valid anymore right, given core will always the objects in the same order, also, we reverted most of the "order agnostic" code, those one that still remain is to fix or add a custom rendering, like `map`, `foreach` and `filter` for example

ryancharris:
@nathaniel-may @marrony Since these tests don't pass without the `UNSTABLE` API version header, I'm going to mark this as WIP until after we've released API v4 👍

PR #337, opened 2020-09-03
FE-767 | AccessProviders API v4

fauna-ryanharris:
### Notes
[Jira Ticket](https://faunadb.atlassian.net/browse/FE-767)

**Note:** We cannot test or verify the functionality here works until API v4 work is done in Core.

This PR updates FQL functions for the API v4 release:
1. Deprecate `Identity` and alias to `CurrentIdentity`
2. Deprecate `HasIdentity` and alias to `HasCurrentIdentity`
3. Add `CurrentToken`
4. Add `HasCurrentToken`
5. Update `CreateAccessProvider`

**TODO:**

- [x] Add tests once Core work is done

### How to test
Run `npm run test`


test/query.test.js
```diff
@@ -2715,6 +2837,91 @@ describe('query', () => {
     }
   })
 
+  // API v4 AccessProvider functions
+  // TODO: Add tests for JWTs using Auth0/Okta
```
nathaniel-may, line 2841: What's the plan for this TODO?

ryancharris: @nathaniel-may I left this stub as part of the work here: https://github.com/fauna/faunadb-js/pull/340#discussion_r490408906

Do we have Auth0 / Okta accounts I can use to test this properly?

nathaniel-may: We don't have them already, but everything is compatible with the free tier of both. We might want to add them as separate integration tests since unlike the other tests here these reach out to a non-fauna service.

ryancharris: @nathaniel-may I queued up a new ticket to address this and will make it my priority in order to get this PR merged in: https://faunadb.atlassian.net/browse/FE-801

danieltodonnell: I agree with this action.  We don't have the infrastructure setup to run tests like this, so we shouldn't block release because of it.

ryancharris: Agreed. In talking with Brecht about the work required to test this, it doesn't seem we're prepared to do so. Also, the integration with Auth0/Okta itself seems to be more of a Core concern than anything.

PR #338, opened 2020-09-04
FE-768 | Update tests for CreateAccessProvider

fauna-ryanharris:
### Notes
[Jira Ticket](https://faunadb.atlassian.net/browse/FE-768)

**NOTE:** #339 needs to be merged first in order for these tests to pass CI

This PR updates the test coverage for `CreateAccessProvider` to reflect the new functionality. Specifically, we now have a `membership` field whose value is an array of Roles.

### How to test
Using the `nightly` version of our Core image locally, you can run `npm run test`

### DIFF
--- a/src/query.js
+++ b/src/query.js
@@ -1032,8 +1032,7 @@ function CreateRole(params) {
  *     - name: A valid schema name
  *     - issuer: A unique string
  *     - jwks_uri: A valid HTTPS URI
- *     - allowed_roles: An optional list of Role refs
- *     - allowed_collections: An optional list of user-defined Collection refs
+ *     - membership: An array of roles
  * @return {Expr}
  */
 function CreateAccessProvider(params) {
@@ -1253,7 +1252,8 @@ function HasIdentity() {
  * See the [docs](https://app.fauna.com/documentation/reference/queryapi#authentication).
  *
  * @return {Expr}
- */  
+ */
+
 function HasCurrentIdentity() {
   arity.exact(0, arguments, HasCurrentIdentity.name)
   return new Expr({ has_current_identity: null })
@@ -1268,7 +1268,7 @@ function CurrentToken() {
   arity.exact(0, arguments, CurrentToken.name)
   return new Expr({ current_token: null })
 }
-  
+
 /**
  * See the [docs](https://app.fauna.com/documentation/reference/queryapi#authentication).
  *
--- a/src/query.js
+++ b/src/query.js
@@ -1032,7 +1032,8 @@ function CreateRole(params) {
  *     - name: A valid schema name
  *     - issuer: A unique string
  *     - jwks_uri: A valid HTTPS URI
- *     - membership: An array of roles
+ *     - membership: An array of role/predicate pairs where the predicate returns a boolean.
+ *                   The array can also contain Role references.
  * @return {Expr}
  */
 function CreateAccessProvider(params) {
--- a/test/query.test.js
+++ b/test/query.test.js
@@ -222,7 +222,7 @@ describe('query', () => {
         new values.Query({
           lambda: '_',
           expr: { let: { x: 1, y: 2 }, in: [{ var: 'x' }, { var: 'y' }] },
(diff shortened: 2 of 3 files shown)


test/query.test.js
```diff
@@ -861,25 +861,63 @@ describe('query', () => {
...
-    expect(provider.jwks_uri).toEqual(fullUri)
+    // Successful instantiation with required fields
+    try {
+      const provider = await adminClient.query(
+        query.CreateAccessProvider({
+          name: providerName,
+          issuer: issuerName,
+          jwks_uri: fullUri,
+          membership: [
+            myNewRole,
```
marrony, line 905: You are passing the role document here, you should pass the reference

fauna-ryanharris: @marrony So it would be `Role(roleOneName)` instead?

test/query.test.js
```diff
@@ -861,25 +861,63 @@ describe('query', () => {
...
+
+    // Successful instantiation with required fields
+    try {
+      const provider = await adminClient.query(
+        query.CreateAccessProvider({
+          name: providerName,
+          issuer: issuerName,
+          jwks_uri: fullUri,
+          membership: [
+            query.Role(`${roleOneName}`),
```
fauna-ryanharris, line 905: @marrony Like this?

marrony: yes.

I don't get why you are using string interpolation here, isn't `roleOneName` a string already? so it would be the same `query.Role(roleOneName)`

fauna-ryanharris, line 905:
```suggestion
            query.Role(roleOneName),
```

test/query.test.js
```diff
@@ -1569,14 +1607,6 @@ describe('query', () => {
     return Promise.all([p1, p2, p3, p4])
   })
 
-  test('contains', () => {
```
marrony, line 1572: I think this should belong to a separate PR

src/query.js
```diff
@@ -1032,8 +1032,7 @@ function CreateRole(params) {
  *     - name: A valid schema name
  *     - issuer: A unique string
  *     - jwks_uri: A valid HTTPS URI
- *     - allowed_roles: An optional list of Role refs
- *     - allowed_collections: An optional list of user-defined Collection refs
+ *     - membership: An array of roles
```
marrony, line 1035: Array of roles doesn't convey the correct purpose of this field. It should say it accepts an object with a pair of role/predicate and that the predicate returns true or false.

test/query.test.js
```diff
@@ -861,25 +861,63 @@ describe('query', () => {
...
+    )
+
+    // Successful instantiation with required fields
+    try {
+      const provider = await adminClient.query(
+        query.CreateAccessProvider({
+          name: providerName,
+          issuer: issuerName,
+          jwks_uri: fullUri,
+          membership: [
```
nathaniel-may, line 904: I like that you used both kinds of membership values here 👍

test/query.test.js
```diff
@@ -861,25 +861,63 @@ describe('query', () => {
...
+          ],
+        })
+      )
+
+      expect(provider.name).toEqual(providerName)
+      expect(provider.issuer).toEqual(issuerName)
+      expect(provider.jwks_uri).toEqual(fullUri)
+      expect(provider.membership).toBeInstanceOf(Array)
+    } catch (error) {
+      expect(error).toBeInstanceOf(errors.BadRequest)
```
nathaniel-may, line 919: Does this mean that if the try block fails, the test will still pass if the error is a BadRequest error? I imagine we'd want to require that the try block is successful.

marrony: yeah, try/catch on tests might not be a good practice, you need to be sure your code will always succeed or fail and do the correct assertions.

fauna-ryanharris: @marrony @nathaniel-may I'll split these tests in two so we can make assertions re: success and failure

nathaniel-may (approved):
LGTM

PR #340, opened 2020-09-17
FE-788 | Add API v4 tests

fauna-ryanharris:
### Notes
[Jira Ticket](https://faunadb.atlassian.net/browse/FE-788)

This PR adds tests for the following functions now that Core has implemented all the new FQL functionality:
- `CurrentToken`
- `HasCurrentToken`
- `CurrentIdentity`
- `HasCurrentIdentity`

### How to test
Run `npm run test` with the latest nightly Docker image

### DIFF
--- a/test/query.test.js
+++ b/test/query.test.js
@@ -2784,7 +2784,7 @@ describe('query', () => {
   })
 
   test('has_current_token returns false when no token present', async () => {
-    const hasCurrentToken = await adminClient.query(query.HasCurrentIdentity())
+    const hasCurrentToken = await client.query(query.HasCurrentToken())
     expect(hasCurrentToken).toBeFalsy()
   })
 
--- a/test/query.test.js
+++ b/test/query.test.js
@@ -2772,11 +2772,12 @@ describe('query', () => {
   })
 
   test('current_token fails when unauthenticated', async () => {
+    const newClient = new Client()
+
     try {
-      const currentToken = await client.query(query.CurrentToken())
-      throw currentToken
+      const currentToken = await newClient.query(query.CurrentToken())
     } catch (err) {
-      expect(err).toBeInstanceOf(errors.BadRequest)
+      expect(err).toBeInstanceOf(errors.Unauthorized)
     }
   })
 
--- a/test/query.test.js
+++ b/test/query.test.js
@@ -222,7 +222,7 @@ describe('query', () => {
         new values.Query({
           lambda: '_',
           expr: { let: { x: 1, y: 2 }, in: [{ var: 'x' }, { var: 'y' }] },
-          api_version: '3',
+          api_version: '4',
         })
       ),
       assertQuery(
@@ -2710,7 +2710,7 @@ describe('query', () => {
 
   // API v4 AccessProvider functions
 
-  test.only('current_identity returns object when authed', async () => {
+  test('current_identity returns object when authed', async () => {
     // Create a new Collection
     const newCollection = await client.query(
       query.Create(collectionRef, { credentials: { password: 'sekrit' } })
@@ -2731,7 +2731,7 @@ describe('query', () => {
     expect(currentIdentity).toBeInstanceOf(Object)
   })
 
-  test.only('current_identity fails when no identity present', async () => {
+  test('current_identity fails when no identity present', async () => {
     try {
       await adminClient.query(query.CurrentIdentity())
     } catch (err) {
@@ -2739,14 +2739,31 @@ describe('query', () => {
     }
   })
 
-  test.only('has_current_identity returns true when identity present', async () => {})
+  test('has_current_identity returns true when identity present', async () => {
+    // Create a new Collection
+    const newCollection = await client.query(
+      query.Create(collectionRef, { credentials: { password: 'sekrit' } })
+    )
 
-  test.only('has_current_identity returns false without identity present', async () => {
+    // Login with permissions for newly created Collection
+    const newLogin = await client.query(
+      query.Login(newCollection['ref'], { password: 'sekrit' })
+    )
+
+    // Get the newly scoped client
+    const newClient = await util.getClient({ secret: newLogin['secret'] })
+
+    // Retrieve the current identity
+    const currentIdentity = await newClient.query(query.CurrentIdentity())
+    expect(currentIdentity).toBeTruthy()
+  })
+
+  test('has_current_identity returns false without identity present', async () => {
     const currentIdentity = await adminClient.query(query.HasCurrentIdentity())
     expect(currentIdentity).toBeFalsy()
   })
 
-  test.only('current_token returns object with AccessProvider', async () => {
+  test('current_token returns object with AccessProvider', async () => {
     const roleOneName = util.randomString('role_one_')
     const issuerName = util.randomString('issuer_')
     const providerName = util.randomString('provider_')
@@ -2781,7 +2798,7 @@ describe('query', () => {
     expect(currentToken).toBeInstanceOf(Object)
   })
 
-  test.only('current_token fails when unauthenticated', async () => {
+  test('current_token fails when unauthenticated', async () => {
     try {
       await adminClient.query(query.CurrentToken())
     } catch (err) {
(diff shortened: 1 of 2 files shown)


test/query.test.js
```diff
@@ -2708,17 +2708,144 @@ describe('query', () => {
     }
   })
 
-  // TODO Create tests once work is done in Core
-  test.skip('current_identity', () => {})
+  // API v4 AccessProvider functions
 
-  // TODO Add test once Core work has been done
-  test.skip('has_current_identity', () => {})
+  test('current_identity returns object when authed', async () => {
```
fauna-ryanharris, line 2713: @nathaniel-may @marrony Are there more cases we need to test for this function?

test/query.test.js
```diff
@@ -2708,17 +2708,144 @@ describe('query', () => {
...
+    const currentIdentity = await newClient.query(query.CurrentIdentity())
+    expect(currentIdentity).toBeTruthy()
+  })
+
+  test('has_current_identity returns false without identity present', async () => {
+    const currentIdentity = await adminClient.query(query.HasCurrentIdentity())
+    expect(currentIdentity).toBeFalsy()
+  })
+
+  test('current_token returns object with AccessProvider', async () => {
```
fauna-ryanharris, line 2766: @nathaniel-may @marrony Are there more cases we need to test for this function?

test/query.test.js
marrony: in theory you should test with JWT tokens too

test/query.test.js
```diff
@@ -2708,17 +2708,144 @@ describe('query', () => {
...
+        name: providerName,
+        issuer: issuerName,
+        jwks_uri: fullUri,
+        membership: [query.Role(roleOneName)],
+      })
+    )
+
+    // Retrieve Ref
+    const currentToken = await adminClient.query(query.CurrentToken())
+    expect(currentToken).toBeInstanceOf(Object)
```
marrony, line 2798: you should assert the `currentToken` is equal the `adminKey.ref`, you can do this by storing the `adminKey` [here](https://github.com/fauna/faunadb-js/blob/master/test/query.test.js#L41) in a global var

test/query.test.js
```diff
@@ -2708,17 +2708,144 @@ describe('query', () => {
...
+    await adminClient.query(
+      query.CreateAccessProvider({
+        name: providerName,
+        issuer: issuerName,
+        jwks_uri: fullUri,
+        membership: [query.Role(roleOneName)],
+      })
+    )
+
+    const hasCurrentToken = adminClient.query(query.HasCurrentToken())
```
marrony, line 2839: same here, if you are using `adminClient` you are using a fauna key, not JWT token, so access provider is useless

test/query.test.js
fauna-ryanharris: Good idea 👍 cc: @danieltodonnell

fauna-ryanharris:
@marrony I hit your notes here. If you're okay with it, we can merge this PR and I'll queue up another ticket to handle the Auth0/Okta tests 👍

test/query.test.js
```diff
@@ -2708,17 +2710,86 @@ describe('query', () => {
...
+  })
+
+  test('current_token returns proper token', async () => {
+    const currentToken = await adminClient.query(query.CurrentToken())
+    expect(currentToken).toEqual(adminKey.ref)
+  })
+
+  test('current_token fails when unauthenticated', async () => {
+    try {
+      await adminClient.query(query.CurrentToken())
```
marrony, line 2776: I have reasons to believe this call is not throwing an exception, then, the `catch` block is not executed, making the test pass, but shouldn't, on try/catch tests like this I usually do this pattern
```
try {
  callTheTestingFunction();

  fail(); //if the testing function didn't throw an error, it will fail anyway
} catch(err) {
  assert on err
}
```

ryancharris: @marrony Good catch! The query was not failing with the `adminClient` and thus not jumping into the `catch` block. I've revised the test to use the normal `client` which successfully throws so I can make the proper assertion in the `catch` 👍

test/query.test.js
ryancharris: I've added a `TODO` for this and we can add these tests once we've set up our Auth0/Okta accounts.

test/query.test.js
```diff
@@ -2708,17 +2710,86 @@ describe('query', () => {
...
+  })
+
+  test('current_token returns proper token', async () => {
+    const currentToken = await adminClient.query(query.CurrentToken())
+    expect(currentToken).toEqual(adminKey.ref)
+  })
+
+  test('current_token fails when unauthenticated', async () => {
+    try {
+      const currentToken = await client.query(query.CurrentToken())
```
marrony, line 2776: using `client` did not change the behavior, `CurrentToken()` is just returning the key ref, regardless the key type you are using, if you don't call something that will actually force the test to fail after the function call, you will never know that the function is actually working, not failing

ryancharris: I am checking that `CurrentToken` works in a separate test above: https://github.com/fauna/faunadb-js/pull/340/files#diff-c2e271250d0fe43a2f2478970f73543eR2769

The `client.query` call itself is causing a failure, which is why the assertion in the `catch` is passing. This is the response I get from Core when logging out `err`:

```
      BadRequest {
        name: 'BadRequest',
        message: 'invalid token',
        description: 'Token metadata is not accessible.',
        requestResult:
         RequestResult {
           method: 'POST',
           path: '',
           query: null,
           requestRaw: '{"current_token":null}',
           requestContent: Expr { raw: [Object] },
           responseRaw:
            '{"errors":[{"position":[],"code":"invalid token","description":"Token metadata is not accessible."}]}',
           responseContent: { errors: [Array] },
           statusCode: 400,
           responseHeaders:
            { connection: 'keep-alive',
              'content-length': '101',
              'content-type': 'application/json;charset=utf-8',
              'x-compute-ops': '1',
              'x-faunadb-build': '20.09.00.nightly20200917-81798af',
              'x-query-bytes-in': '22',
              'x-query-bytes-out': '101',
              'x-query-time': '4',
              'x-read-ops': '0',
              'x-storage-bytes-read': '0',
              'x-storage-bytes-write': '0',
              'x-txn-retries': '0',
              'x-txn-time': '1600434387785000',
              'x-write-ops': '0' },
           startTime: 1600434387779,
           endTime: 1600434387786 } }
```

marrony: I see why this is returning the error, but this is not related to the fact it's unauthenticated. It's because the `client` is using a key that points to a sub database, however, the key lives in the parent database (because it was created with the [rootClient](https://github.com/fauna/faunadb-js/blob/d44f3c956eb2c1e3efc802e9d55c65cd65b87e22/test/util.js#L129-L144)), so that key is not accessible from the db it's pointing.

BrunoQuaresma (commented):
LGTM. I would like to wait to approve until have the conflicts solved.

ryancharris:
> LGTM. I would like to wait to approve until have the conflicts solved.

@BrunoQuaresma Just resolved the conflict 👍

BrunoQuaresma (approved):
LGTM. Here I have the same question related to the tests on https://github.com/fauna/faunadb-js/pull/330#issuecomment-696713798

ryancharris:
> LGTM. Here I have the same question related to the tests on [#330 (comment)](https://github.com/fauna/faunadb-js/pull/330#issuecomment-696713798)

@marrony @nathaniel-may Thoughts re: Bruno's question?

test/query.test.js
```diff
@@ -2835,17 +2837,88 @@ describe('query', () => {
...
+    expect(currentIdentity).toEqual(newLogin.instance)
+    expect(currentIdentity).toBeInstanceOf(Object)
+  })
+
+  test('current_identity fails when no identity present', async () => {
+    try {
+      await adminClient.query(query.CurrentIdentity())
+    } catch (err) {
+      expect(err).toBeInstanceOf(errors.BadRequest)
+    }
```
marrony, line 2869: in my opinion this test is still error prune, it can pass, even producing wrong results if the `CurrentIdentity()` doesn't throw an error, my suggestion is to do something:
```suggestion
    try {
      await adminClient.query(query.CurrentIdentity())
      
      throw new Error('Should not reached here')
    } catch (err) {
      expect(err).toBeInstanceOf(errors.BadRequest)
    }
```

PR #344, opened 2020-09-24
Minor: mark internal functions as private for JSDoc

erickpintor:
Tagging private functions so they don't appear in the library documentation.


BrunoQuaresma (approved):
LGTM.

PR #356, opened 2020-10-15
FE-808 | Add TS types for streaming

ryancharris:
### Notes
[Jira Ticket](https://faunadb.atlassian.net/browse/FE-808)

Eric Wyne reported that our TS types were not updated when adding streaming support. This PR is intended to do that 👍

### DIFF
--- a/src/stream.js
+++ b/src/stream.js
@@ -370,7 +370,7 @@ EventDispatcher.prototype.dispatch = function(event) {
  *   * 'action': The action type
  *   * 'document': The document's data
  *   * 'diff': The difference between 'document' and 'prev'
- *   * 'prev': The event's old data
+ *   * 'prev': The event's previous data
  */
 
 /**
--- a/src/stream.js
+++ b/src/stream.js
@@ -340,7 +340,7 @@ EventDispatcher.prototype.dispatch = function(event) {
  *
  * @event module:stream~Subscription#snapshot
  * @type {object}
- * @property {string} event='snapshot'
+ * @property {string} type='snapshot'
  *   The event type.
  * @property {number} txn
  *   The event's transaction timestamp.
--- a/src/stream.js
+++ b/src/stream.js
@@ -92,14 +92,14 @@ StreamClient.prototype.snapshot = function() {
     .query(q.Get(self._query))
     .then(function(doc) {
       self._onEvent({
-        event: 'snapshot',
-        data: doc,
+        type: 'snapshot',
+        event: doc,
       })
     })
     .catch(function(error) {
       self._onEvent({
-        event: 'error',
-        data: error,
+        type: 'error',
+        event: error,
       })
     })
 }
@@ -451,7 +451,7 @@ function StreamAPI(client) {
     var streamClient = new StreamClient(client, expression, options, onEvent)
 
     function onEvent(event) {
-      switch (event.event) {
+      switch (event.type) {
         case 'start':
           dispatcher.dispatch(event)
           streamClient.snapshot()
--- a/src/stream.js
+++ b/src/stream.js
@@ -475,7 +475,7 @@ function StreamAPI(client) {
       dispatcher.dispatch(snapshotEvent)
       for (var i = 0; i < buffer.length; i++) {
         var bufferedEvent = buffer[i]
-        if (bufferedEvent.txnTS > snapshotEvent.data.ts) {
+        if (bufferedEvent.txn > snapshotEvent.event.ts) {
           dispatcher.dispatch(bufferedEvent)
         }
       }
--- a/src/stream.js
+++ b/src/stream.js
@@ -354,7 +354,7 @@ EventDispatcher.prototype.dispatch = function(event) {
  *
  * @event module:stream~Subscription#error
  * @type {object}
- * @property {string} event='error'
+ * @property {string} type='error'
  *   The event type.
  * @property {?number} txn
  *   The event's transaction timestamp.
--- a/src/stream.js
+++ b/src/stream.js
@@ -325,7 +325,7 @@ EventDispatcher.prototype.dispatch = function(event) {
  *
  * @event module:stream~Subscription#history_rewrite
  * @type {object}
- * @property {string} event='history_rewrite'
+ * @property {string} type='history_rewrite'
  *   The event type.
  * @property {number} txn
  *   The event's transaction timestamp.
--- a/src/stream.js
+++ b/src/stream.js
@@ -344,7 +344,7 @@ EventDispatcher.prototype.dispatch = function(event) {
  *   The event type.
  * @property {number} txn
  *   The event's transaction timestamp.
- * @property {object} data
+ * @property {object} event
  *   The event's data.
  */
 
--- a/src/stream.js
+++ b/src/stream.js
@@ -297,11 +297,11 @@ EventDispatcher.prototype.dispatch = function(event) {
  *
  * @event module:stream~Subscription#start
  * @type {object}
- * @property {string} event='start'
+ * @property {string} type='start'
  *   The event type.
- * @property {number} txnTS
+ * @property {number} txn
  *   The event's transaction timestamp.
- * @property {module:values~FaunaTime} data
+ * @property {module:number} event
  *   The stream start timestamp.
  */
 
@@ -311,11 +311,11 @@ EventDispatcher.prototype.dispatch = function(event) {
  *
  * @event module:stream~Subscription#version
  * @type {object}
- * @property {string} event='version'
+ * @property {string} type='version'
  *   The event type.
- * @property {number} txnTS
+ * @property {number} txn
  *   The event's transaction timestamp.
- * @property {object} data
+ * @property {object} event
  *   The event's data.
  */
 
@@ -327,9 +327,9 @@ EventDispatcher.prototype.dispatch = function(event) {
  * @type {object}
  * @property {string} event='history_rewrite'
  *   The event type.
- * @property {number} txnTS
+ * @property {number} txn
  *   The event's transaction timestamp.
- * @property {object} data
+ * @property {object} event
  *   The event's data.
  */
 
@@ -340,9 +340,9 @@ EventDispatcher.prototype.dispatch = function(event) {
  *
  * @event module:stream~Subscription#snapshot
  * @type {object}
- * @property {string} event='version'
+ * @property {string} event='snapshot'
  *   The event type.
- * @property {number} txnTS
+ * @property {number} txn
  *   The event's transaction timestamp.
  * @property {object} data
  *   The event's data.
@@ -356,23 +356,21 @@ EventDispatcher.prototype.dispatch = function(event) {
  * @type {object}
  * @property {string} event='error'
  *   The event type.
- * @property {?number} txnTS
+ * @property {?number} txn
  *   The event's transaction timestamp.
- * @property {Error} data
+ * @property {Error} event
  *   The underlying error.
  */
 
 /**
  * @typedef {Object} Options
- * @property {string[]} [fields=['ref', 'ts', 'action', 'new']]
+ * @property {string[]} [fields=['action', 'document', 'diff', 'prev']]
  *   The fields event fields to opt-in during stream subscription. Possible
  *   options:
- *   * 'ref': The event's document reference;
- *   * 'ts': The event's valid time;
- *   * 'new': The event's new data;
- *   * 'old': The event's old data;
- *   * 'diff': The event's data difference;
- *   * 'action': The event's action.
+ *   * 'action': The action type
+ *   * 'document': The document's data
+ *   * 'diff': The difference between 'document' and 'prev'
+ *   * 'prev': The event's old data
  */
 
 /**
--- a/src/stream.js
+++ b/src/stream.js
@@ -159,8 +159,8 @@ StreamClient.prototype.subscribe = function() {
     var events = json.parseJSONStreaming(data)
 
     events.forEach(function(event) {
-      if (event.txnTS !== undefined) {
-        self._client.syncLastTxnTime(event.txnTS)
+      if (event.txn !== undefined) {
+        self._client.syncLastTxnTime(event.txn)
       }
 
       if (event.event === 'error') {
@@ -176,8 +176,8 @@ StreamClient.prototype.subscribe = function() {
     // close() on a Subscription. There's no need to relay this event back up.
     if (error.name !== 'AbortError') {
       self._onEvent({
-        event: 'error',
-        data: error,
+        type: 'error',
+        event: error,
       })
     }
   }
@@ -280,13 +280,13 @@ EventDispatcher.prototype.on = function(type, callback) {
  * @param {Object} event The event.
  */
 EventDispatcher.prototype.dispatch = function(event) {
-  var listeners = this._listeners[event.event]
+  var listeners = this._listeners[event.type]
   if (!listeners) {
     return
   }
 
   for (var i = 0; i < listeners.length; i++) {
-    listeners[i].call(null, event.data, event)
+    listeners[i].call(null, event.event, event)
   }
 }
 
--- a/src/types/Client.d.ts
+++ b/src/types/Client.d.ts
@@ -2,7 +2,7 @@ import Expr from './Expr'
 import { ExprArg } from './query'
 import PageHelper from './PageHelper'
 import RequestResult from './RequestResult'
-import { ClientStreamOptions, Subscription } from './Stream'
+import { Subscription } from './Stream'
 
 export interface ClientConfig {
   secret: string
@@ -26,5 +26,5 @@ export default class Client {
   query<T = object>(expr: ExprArg, options?: QueryOptions): Promise<T>
   paginate(expr: Expr, params?: object, options?: QueryOptions): PageHelper
   ping(scope?: string, timeout?: number): Promise<string>
-  stream(expr: Expr, options?: ClientStreamOptions): Subscription
+  stream(expr: Expr, options?: { fields: string[] }): Subscription
 }
--- a/src/types/Client.d.ts
+++ b/src/types/Client.d.ts
@@ -14,7 +14,6 @@ export interface ClientConfig {
   keepAlive?: boolean
   headers?: { [key: string]: string | number }
   fetch?: typeof fetch
-  stream?: Subscription
 }
 
 export interface QueryOptions {
--- a/src/types/Client.d.ts
+++ b/src/types/Client.d.ts
@@ -2,7 +2,7 @@ import Expr from './Expr'
 import { ExprArg } from './query'
 import PageHelper from './PageHelper'
 import RequestResult from './RequestResult'
-import { Subscription } from './Stream'
+import { StreamSubscription } from './Stream'
 
 export interface ClientConfig {
   secret: string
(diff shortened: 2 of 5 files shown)


ryancharris:
@BrunoQuaresma @rodrimaia I was looking at the TS docs and it seems we can auto-generate our `.d.ts` files by updating our `.tsconfig` file: https://www.typescriptlang.org/docs/handbook/declaration-files/dts-from-js.html

What do you think?

src/types/Stream.d.ts
```diff
@@ -1,10 +1,27 @@
...
+
+export interface StreamClient {
+  client: Client
+  expression: ExprArg
+  options: Options
+  onEvent: Function
+}
+
+export interface Options {
+  ref: string
```
BrunoQuaresma, line 21: Is this a string or a Ref object?

src/types/Stream.d.ts
```diff
@@ -1,10 +1,27 @@
...
 
 export interface EventDispatcher {
   allowedEvents: string[]
 }
+
+export interface StreamClient {
+  client: Client
+  expression: ExprArg
+  options: Options
+  onEvent: Function
```
BrunoQuaresma, line 17: This is too generic, what should we expect here? what arguments and response?

src/types/Stream.d.ts
```diff
@@ -1,10 +1,27 @@
...
+import { ExprArg } from './query'
 
 export interface Subscription {
-  client: Client
+  client: StreamClient
   dispatcher: EventDispatcher
 }
 
 export interface EventDispatcher {
   allowedEvents: string[]
```
BrunoQuaresma, line 10: This looks too generic as well, we should say specifically the events.

src/types/Stream.d.ts
```diff
@@ -1,10 +1,27 @@
...
+  onEvent: Function
+}
+
+export interface Options {
+  ref: string
+  ts: string
+  new: string
+  old: string
+  diff: string
+  action: string
```
BrunoQuaresma, line 26: I also think this is too generic. If the action is pre-defined, we should put the specific values.

BrunoQuaresma (commented):
Some notes.

src/types/Stream.d.ts
ecwyne: This may help https://cjam.github.io/typesafe-event-emitter-in-typescript/

src/types/Client.d.ts
```diff
@@ -13,6 +14,7 @@ export interface ClientConfig {
   keepAlive?: boolean
   headers?: { [key: string]: string | number }
   fetch?: typeof fetch
+  stream?: Subscription
```
erickpintor, line 17: I think this filed is misplaced. There isn't a `stream` field in the client config object as fas as I know.

src/types/Client.d.ts
```diff
@@ -25,4 +27,5 @@ export default class Client {
   query<T = object>(expr: ExprArg, options?: QueryOptions): Promise<T>
   paginate(expr: Expr, params?: object, options?: QueryOptions): PageHelper
   ping(scope?: string, timeout?: number): Promise<string>
+  stream(expr: Expr, options?: object): Subscription
```
erickpintor, line 30: Could you also type this `options` argument? It has a single field called `fields` and it is a list of strings.

src/types/Stream.d.ts
erickpintor: I don't think it's worth documenting the internal types. Users don't have access to it.

src/types/Stream.d.ts
```diff
@@ -0,0 +1,27 @@
+import Client from './Client'
+import { ExprArg } from './query'
+
+export interface Subscription {
+  client: StreamClient
+  dispatcher: EventDispatcher
```
erickpintor, line 6: This fields don't exist, their `_` prefixed counterparts are private to the subscription class.

erickpintor: I think we should document its public interface instead: `on`, `start`, and `close`. Also, the most useful type annotation will probably by the events themselves.

src/types/Stream.d.ts
erickpintor: This class is internal. We should focus on the public interface, IMO.

src/types/Stream.d.ts
```diff
@@ -0,0 +1,27 @@
...
+}
+
+export interface StreamClient {
+  client: Client
+  expression: ExprArg
+  options: Options
+  onEvent: Function
+}
+
+export interface Options {
```
erickpintor, line 20: What is this object for?

erickpintor (changes requested):

src/types/Stream.d.ts
```diff
@@ -0,0 +1,10 @@
+import Client from './Client'
+import { ExprArg } from './query'
+
+export interface Subscription {
+  on: (type: SubscriptionEvents, callback: Function) => this
+  start: () => this
+  close: () => this
+}
+
+type SubscriptionEvents = 'ref' | 'ts' | 'new' | 'old' | 'diff' | 'action'
```
erickpintor, line 10: These are fields that come with the "version" event. The event types are: "start", "version", "history_rewrite", and "error". The list of events can be found in the docs here: https://github.com/fauna/faunadb-js/blob/master/src/stream.js#L289-L372

Here is the full list of events and their fields for convenience:

```
event: start
  fields:
    - event: the event name
    - txnTS: the event transaction timestamp
    - data: the stream start timestamp

event: version
  fields:
    - event: the event name
    - txnTS: the event transaction timestamp
    - data:
      - ref: the document's ref
      - ts: the document's valid timestamp
      - action: the action that triggered the event
      - new: the document's new data
      - old: the document's old data (optional)
      - diff: the diff between old and new data (optional)

event: history_rewrite
  fields:
    - event: the event name
    - txnTS: the event transaction timestamp
    - data:
      - ref: the document's ref
      - ts: the document's valid timestamp
      - action: the action that triggered the event

event: error
  fields:
    - event: the event name
    - txnTS: the event transaction timestamp
    - data:
      - code: the error code
      - description: the error description
```

src/types/Client.d.ts
```diff
@@ -25,4 +26,5 @@ export default class Client {
   query<T = object>(expr: ExprArg, options?: QueryOptions): Promise<T>
   paginate(expr: Expr, params?: object, options?: QueryOptions): PageHelper
   ping(scope?: string, timeout?: number): Promise<string>
+  stream(expr: Expr, options?: { fields: string[] }): Subscription
```
erickpintor, line 29: `fields` itself is also optional.

src/types/Stream.d.ts
```diff
@@ -0,0 +1,40 @@
+import Client from './Client'
+import { ExprArg } from './query'
+import { FaunaTime, Ref } from './values'
+
+export interface Subscription {
+  on: (type: SubscriptionEvents, callback: Function) => this
+  start: () => this
+  close: () => this
```
erickpintor, line 8: `close` doesn't return anything.

src/types/Stream.d.ts
```diff
@@ -0,0 +1,40 @@
...
+
+type StreamStartEventData = FaunaTime
+
+type StreamVersionEventData = {
+  ref: Ref
+  ts: number
+  action: string
+  new: object
+  old: object
+  diff: string
```
erickpintor, line 26: `diff` is also an object.

src/types/Stream.d.ts
```diff
@@ -0,0 +1,40 @@
...
+}
+
+type StreamErrorEventData = {
+  code: number
```
erickpintor, line 30: `code` is a string.

src/types/Stream.d.ts
```diff
@@ -0,0 +1,40 @@
...
+  description: string
+}
+
+type StreamHistoryRewriteEventData = {
+  ref: Ref
+  ts: number
+  action: string
+}
+
+type SubscriptionEvents = 'start' | 'version' | 'history_rewrite' | 'error'
```
erickpintor, line 40: Question: is it possible in TS to tie the literal string with the event type received?

ecwyne: Found this. https://github.com/bterlson/strict-event-emitter-types

src/types/Stream.d.ts
```diff
@@ -0,0 +1,53 @@
...
+  | StreamStartSubscription
+  | StreamVersionSubscription
+  | StreamErrorSubscription
+  | StreamHistoryRewriteSubscription
+
+interface Subscription {
+  start: () => this
+  close: () => void
+  txnTS: number
+}
```
erickpintor, line 15: The subscription itself does not contain data. The previews version had the correct type signature:

```js
interface Subscription {
  on: (type: SubscriptionEvents, callback: Function) => this
  start: () => this
  close: () => this
}
```

What I wanted to try is to tie the `SubscriptionEvents` literal string with the type of the `callback` parameter.

For example:

```js
client.stream(ref).on('start', ts => ...)
//                             ^ 
// I wanted typescript to understand that this is a FaunaTime
```

@BrunoQuaresma mentioned earlier in the project that maybe that's possible. I don't know much about TS, though. I'm okay if that's not feasible.

ecwyne:
I believe you're looking for something like this.

```ts
// Fill out these handler type signatures
type SubscriptionEventHandlers = {
    start: (ts: FaunaTime, e: {event: 'start', txnTs: number, data: FaunaTime}) => void;
    error: () => void;
    version: () => void;
    history_rewrite: () => void;
    snapshot: () => void;
};

interface Subscription {
    on: <T extends keyof SubscriptionEventHandlers>(
        type: T,
        callback: SubscriptionEventHandlers[T],
    ) => this;
    start: () => this;
    close: () => void;
}
```

src/types/Stream.d.ts
ecwyne: https://github.com/fauna/faunadb-js/pull/356#issuecomment-713202337

ryancharris:
@erickpintor What fields am I missing here? https://github.com/fauna/faunadb-js/pull/356/files#diff-21ff9a7035770e3d38455382a3b21b2720ee2fc5894cd9ee15d2502201b2f92cR66

src/types/Client.d.ts
```diff
@@ -25,4 +26,5 @@ export default class Client {
   query<T = object>(expr: ExprArg, options?: QueryOptions): Promise<T>
   paginate(expr: Expr, params?: object, options?: QueryOptions): PageHelper
   ping(scope?: string, timeout?: number): Promise<string>
+  stream(expr: Expr, options?: { fields?: string[] }): Subscription
```
erickpintor, line 29: Not a requirement. More of a question: is it possible to specify the valid values for this array?

src/types/Stream.d.ts
```diff
@@ -0,0 +1,69 @@
...
+  on: <T extends keyof SubscriptionEventHandlers>(
+    type: T,
+    callback: SubscriptionEventHandlers[T]
+  ) => this
+  start: () => this
+  close: () => void
+}
+
+type SubscriptionEventHandlers = {
+  start: (type: 'start', txn: values.FaunaTime, event: values.FaunaTime) => void
```
erickpintor, line 15: Both txn and event are no longer `FaunaTime` instances. They are numbers now.

erickpintor: The same thing goes for for all `txn` and `ts` fields in this type annotation.

src/types/Stream.d.ts
```diff
@@ -0,0 +1,69 @@
...
+        ref: values.Ref
+        ts: values.FaunaTime
+        data: object
+      }
+      prev: {
+        ref: values.Ref
+        ts: values.FaunaTime
+        data?: object
+      }
+    }
```
erickpintor, line 44: I appreciate the careful typing here but, I think this are better off as just `object` as their rendering are highly dynamic and dependent on the document being streamed. For instance, data is never guaranteed to be present. Once we enable schema documents, they will have other properties, like `name`, `roles`, etc. Trying to cover all possibilities at this level will be quite tedious and error prone.

erickpintor: This is the same for other `document` fields in this type annotation.

ecwyne: If I can provide some user feedback. I would much rather have incomplete/error-prone types over simply `any` or `object`. TS types provide not only type-safety for refactoring but also serve as a main source for documentation when learning/exploring new parts of the api.

erickpintor: As a user, can you work around incomplete types in TS? If so, I'd say mark data as optional and leave it as is.

ryancharris:
@erickpintor Any idea why this might be passing in one Preview environment, but not the other?

src/stream.js
```diff
@@ -327,9 +327,9 @@ EventDispatcher.prototype.dispatch = function(event) {
  * @type {object}
  * @property {string} event='history_rewrite'
```
erickpintor, line 328:
```suggestion
 * @property {string} type='history_rewrite'
```

src/stream.js
```diff
@@ -340,9 +340,9 @@ EventDispatcher.prototype.dispatch = function(event) {
  *
  * @event module:stream~Subscription#snapshot
  * @type {object}
- * @property {string} event='version'
+ * @property {string} event='snapshot'
```
erickpintor, line 343:
```suggestion
 * @property {string} type='snapshot'
```

src/stream.js
```diff
@@ -340,9 +340,9 @@ EventDispatcher.prototype.dispatch = function(event) {
...
  *   The event type.
- * @property {number} txnTS
+ * @property {number} txn
  *   The event's transaction timestamp.
  * @property {object} data
```
erickpintor, line 347:
```suggestion
 * @property {object} event
```

src/stream.js
```diff
@@ -356,23 +356,21 @@ EventDispatcher.prototype.dispatch = function(event) {
  * @type {object}
  * @property {string} event='error'
```
erickpintor, line 357:
```suggestion
 * @property {string} type='error'
```

src/stream.js
```diff
@@ -356,23 +356,21 @@ EventDispatcher.prototype.dispatch = function(event) {
...
- *   * 'ref': The event's document reference;
- *   * 'ts': The event's valid time;
- *   * 'new': The event's new data;
- *   * 'old': The event's old data;
- *   * 'diff': The event's data difference;
- *   * 'action': The event's action.
+ *   * 'action': The action type
+ *   * 'document': The document's data
+ *   * 'diff': The difference between 'document' and 'prev'
+ *   * 'prev': The event's old data
```
erickpintor, line 373:
```suggestion
 *   * 'prev': The event's previews data
```

src/types/Stream.d.ts
```diff
@@ -0,0 +1,50 @@
...
+    event: {
+      action: 'history_rewrite'
+      document: object
+    }
+  ) => void
+  snapshot: (
+    type: 'snapshot',
+    txn: number,
+    event: {
+      action: 'snapshot'
```
erickpintor, line 46: A snapshot has no action. The event itself is the document.

erickpintor:
> Any idea why this might be passing in one Preview environment, but not the other?

We gotta check case by case. I think the node version has something to do with it:

```
TypeError: results.flat is not a function

      49 |   }
      50 | 
    > 51 |   return results.flat()
```

Also, preview has the new event format, while I don't think stable has it yet.

src/stream.js
ryancharris: @erickpintor I'm a bit confused by this annotation -- I thought the `prev` field represented the old document data?

src/types/Stream.d.ts
ryancharris: @erickpintor In that case, would `event: object` be the better definition? Or is there an `event.document` field that should be an `object`?

ryancharris:
> Also, preview has the new event format, while I don't think stable has it yet.

@erickpintor What implications will this have?

ryancharris:
> We gotta check case by case. I think the node version has something to do with it:

@erickpintor You are correct, sir! It seems the tests pass whenever we're using Node v11+. Locally (using NVM), I ran the tests with these various Node versions:

```
Tests fail with these Node versions
- 10.18.0
- 10.23.0

Tests pass with these Node versions
- 11.0.0
- 11.15.0
- 12.19.0
- 14.2.0
```

ryancharris:
@erickpintor This is the TLDR of why the tests are passing/failing:

```
** Our environments are using the following Node versions:
These all fail due to Node version issues
- core-stable-10: Node 10
- core-nightly-10: Node 10
- core-preview-10: Node 10

These fail due to feature functionality
- core-stable-11: Node 11
- core-nightly-11: Node 11

This passes since it's the right Node version and has streaming enabled
- core-preview-11: Node 11
```

erickpintor:
Here are my findings:

* The JS driver has code that is broken in node 10 (we gotta fix that);
* Stable docker image has no stream api enabled, hence stream tests will fail with it;
* The night build has the new event’s format, where the branch doesn’t. Therefore, it will fail with it;
* The preview docker image has the old event’s format, which matches with the branch’s code so, that’s why it’s the only scenario passing.

src/types/Stream.d.ts
erickpintor: `event: object` should it. There's no `document` field in it.

src/stream.js
erickpintor: It is the old documents data. I thought using the word preview would be more consistent with the field named `prev`. I'm fine with "old" too.

test/stream.test.js
```diff
@@ -92,7 +92,7 @@ describe('StreamAPI', () => {
           client.query(q.Update(doc.ref, {}))
         })
         .on('version', (_, event) => {
-          expect(client.getLastTxnTime()).toEqual(event.txnTS)
+          expect(client.getLastTxnTime()).toEqual(event.event.document.ts)
```
erickpintor, line 95: The transaction time should be tracked against the `event.txn` field.
```suggestion
          expect(client.getLastTxnTime()).toEqual(event.txn)
```

src/stream.js
ryancharris: @erickpintor Ah, okay. I'll go with `previous` cause that sounds more semantic 👍

test/stream.test.js
```diff
@@ -235,7 +234,7 @@ describe('StreamAPI', () => {
         .start()
     })
 
-    test('filter buffered events prior to snapshot', done => {
+    test.skip('filter buffered events prior to snapshot', done => {
```
erickpintor, line 237: This patch fixes the test:

```diff
diff --git a/src/stream.js b/src/stream.js
index 00d4e36..f5828d9 100644
--- a/src/stream.js
+++ b/src/stream.js
@@ -475,7 +475,7 @@ function StreamAPI(client) {
       dispatcher.dispatch(snapshotEvent)
       for (var i = 0; i < buffer.length; i++) {
         var bufferedEvent = buffer[i]
-        if (bufferedEvent.txnTS > snapshotEvent.data.ts) {
+        if (bufferedEvent.txn > snapshotEvent.event.ts) {
           dispatcher.dispatch(bufferedEvent)
         }
       }
```

test/stream.test.js
ryancharris: @erickpintor That's what I originally thought, but the test wasn't consistently passing 🤔

test/stream.test.js
ryancharris: @erickpintor Of course! I don't know how I missed that 😅 I was looking at this file for way too long.

erickpintor (approved):
LGTM

test/stream.test.js
```diff
@@ -59,10 +59,10 @@ describe('StreamAPI', () => {
     test('can listen to events', done => {
       stream = client
         .stream(doc.ref)
-        .on('start', (ts, event) => {
-          expect(event.event).toEqual('start')
-          expect(event.txnTS).toBeDefined()
-          expect(event.data).toEqual(ts)
+        .on('start', (_, event) => {
```
danieltodonnell, line 62: I see the _ was the part you were stuck on to get the test running. nice.

danieltodonnell (approved):
Didn't run it but code LGTM

nathaniel-may (approved):

PR #358, opened 2020-10-23
handle multiple json objects in a message

BrunoQuaresma:
JIRA Ticket: https://faunadb.atlassian.net/browse/ENG-2374

### DIFF
--- a/src/_json.js
+++ b/src/_json.js
@@ -16,6 +16,36 @@ function parseJSON(json) {
   return JSON.parse(json, json_parse)
 }
 
+function parseJSONStreaming(json) {
+  var results = []
+
+  try {
+    results.push(parseJSON(json))
+  } catch (error) {
+    var position
+
+    /** JSON parse error message is throw by the browser */
+    if (error.message.match(/JSON.parse/g)) {
+      var matchResult = error.message.match(/column ([0-9+]*)/)
+      position = matchResult[1] - 1
+      /** Unexpected token message is throw by NodeJS */
+    } else if (error.message.match(/Unexpected token/g)) {
+      var matchResult = error.message.match(/at position ([0-9+]*)$/)
+      position = matchResult[1]
+    } else {
+      throw error
+    }
+
+    var left = json.slice(0, position)
+    var right = json.slice(position)
+
+    results.push(parseJSONStreaming(left))
+    results.push(parseJSONStreaming(right))
+  }
+
+  return results.flat()
+}
+
 function json_parse(_, val) {
   if (typeof val !== 'object' || val === null) {
     return val
@@ -50,4 +80,5 @@ function json_parse(_, val) {
 module.exports = {
   toJSON: toJSON,
   parseJSON: parseJSON,
+  parseJSONStreaming: parseJSONStreaming,
 }
--- a/src/_json.js
+++ b/src/_json.js
@@ -16,6 +16,11 @@ function parseJSON(json) {
   return JSON.parse(json, json_parse)
 }
 
+/**
+ * This function allow us to parse responses as streaming
+ * currently, this is not allowed by Firefox only
+ * https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API
+ */
 function parseJSONStreaming(json) {
   var results = []
 
--- a/src/_json.js
+++ b/src/_json.js
@@ -24,12 +24,12 @@ function parseJSONStreaming(json) {
   } catch (error) {
     var position
 
-    /** JSON parse error message is throw by the browser */
-    if (error.message.match(/JSON.parse/g)) {
+    /** NodeJS and Chrome Syntax error message */
+    if (error.message.indexOf('JSON.parse') > -1) {
       var matchResult = error.message.match(/column ([0-9+]*)/)
       position = matchResult[1] - 1
-      /** Unexpected token message is throw by NodeJS */
-    } else if (error.message.match(/Unexpected token/g)) {
+      /** Firefox Syntax error message */
+    } else if (error.message.indexOf('Unexpected token') > -1) {
       var matchResult = error.message.match(/at position ([0-9+]*)$/)
       position = matchResult[1]
     } else {
--- a/src/stream.js
+++ b/src/stream.js
@@ -210,7 +210,7 @@ StreamClient.prototype.subscribe = function() {
             .read()
             .then(function process(msg) {
               if (!msg.done) {
-                onData(decoder.decode(msg.value))
+                onData(decoder.decode(msg.value, { stream: true }))
                 return pump()
               }
             })
--- a/src/stream.js
+++ b/src/stream.js
@@ -156,15 +156,19 @@ StreamClient.prototype.subscribe = function() {
   }
 
   function onData(data) {
-    var event = json.parseJSON(data)
-    if (event.txnTS !== undefined) {
-      self._client.syncLastTxnTime(event.txnTS)
-    }
-    if (event.event === 'error') {
-      onError(new errors.StreamErrorEvent(event))
-    } else {
-      self._onEvent(event)
-    }
+    var events = json.parseJSONStreaming(data)
+
+    events.forEach(function(event) {
+      if (event.txnTS !== undefined) {
+        self._client.syncLastTxnTime(event.txnTS)
+      }
+
+      if (event.event === 'error') {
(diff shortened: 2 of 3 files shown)


src/_json.js
```diff
@@ -16,6 +16,36 @@ function parseJSON(json) {
...
+
+  try {
+    results.push(parseJSON(json))
+  } catch (error) {
+    var position
+
+    /** JSON parse error message is throw by the browser */
+    if (error.message.match(/JSON.parse/g)) {
+      var matchResult = error.message.match(/column ([0-9+]*)/)
+      position = matchResult[1] - 1
```
faunaee, line 30: We should avoid multiple regex operations where possible; this code could get called very frequently.

```suggestion
    var capture
    if (capture = error.message.match(/JSON.parse: unexpected non-whitespace.*column (\d+)/)) {
      position = capture[1] - 1
```

src/_json.js
```diff
@@ -16,6 +16,36 @@ function parseJSON(json) {
...
+      /** Unexpected token message is throw by NodeJS */
+    } else if (error.message.match(/Unexpected token/g)) {
+      var matchResult = error.message.match(/at position ([0-9+]*)$/)
+      position = matchResult[1]
```
faunaee, line 34: Use the same strategy here.

Note that your capture group syntax seems off here. `\d` matches digits. I'm not sure why your character array includes `+`, nor why would accept 0 or more digits.

The capture group in my suggestion above seems better.

src/_json.js
```diff
@@ -16,6 +16,36 @@ function parseJSON(json) {
...
+    } else {
+      throw error
+    }
+
+    var left = json.slice(0, position)
+    var right = json.slice(position)
```
faunaee, line 40: `slice` is for arrays. Why don't we use `substr` here? Then you don't have to do position math (above).

faunaee (changes requested):
A few comments on the implementation.

The new `parseJSONStreaming` function should definitely have unit tests covering single and multiple JSON objects.

src/_json.js
danieltodonnell: In a browser at least, `.indexOf()` is 10 times faster than a regex, so I would suggest this instead.  We test FOR the error quickly, then use the more expensive operation to get the position if it's happened

if (error.message.indexOf('JSON.parse')) {
      var matchResult = error.message.match(/column ([0-9+]*)/)

src/_json.js
danieltodonnell: Likewise, i think indexOf will be more efficient.

src/_json.js
danieltodonnell: https://www.measurethat.net/Benchmarks/Show/975/11/regex-vs-indexof-vs-startswith-vs-substr#latest_results_block

src/_json.js
danieltodonnell: Ewan is correct about the regex.. change it to `/at position ([0-9]+)/`

danieltodonnell (changes requested):

src/_json.js
danieltodonnell: This regex is incorrect as well

use `/column ([0-9]+)/`

BrunoQuaresma: > This regex is incorrect as well
> 
> use `/column ([0-9]+)/`

It is not. `column ([0-9]+)` didn't work. 😕

BrunoQuaresma: Ok, I'm going to use `indexOf` to check the error.

BrunoQuaresma:
> The new `parseJSONStreaming` function should definitely have unit tests covering single and multiple JSON objects.

I totally agree with this 😄 . I just wanted to collect some feedback first.

src/_json.js
BrunoQuaresma: The same of before, `/at position ([0-9]+)/` didn't work.

src/_json.js
BrunoQuaresma: I think for this case there is no diff, and String also has a `slice()` method.

https://stackoverflow.com/questions/2243824/what-is-the-difference-between-string-slice-and-string-substring

src/_json.js
faunaee: @danieltodonnell How is the regex that I suggested incorrect? I tested it against the error message quoted in the issue description. I might be overlooking regex support in browsers that the driver supports, so I'm curious what I'm missing.

erickpintor: Are we sure the message is stable across browsers + node js versions?

erickpintor (commented):
For safely, we should also set the text decoder to streaming mode.

src/_json.js
BrunoQuaresma: @erickpintor I was only able to catch these 2 messages but I'm not 100% sure these are the only possible messages.

BrunoQuaresma:
Interesting discovery: Firefox JS error messages are diff from Chrome that are diff from Safari and go on.

Screen Shot 2020-10-26 at 09 16 16
Screen Shot 2020-10-26 at 09 14 38
Screen Shot 2020-10-26 at 09 14 24

I think we can move forward with the current solution since this issue only happens, from what we know, in Firefox.

erickpintor:
https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API

image

It does seem like a firefox only issue. Would you mind adding some notes in the code for it?

src/stream.js
```diff
@@ -156,15 +156,19 @@ StreamClient.prototype.subscribe = function() {
...
-      self._client.syncLastTxnTime(event.txnTS)
-    }
-    if (event.event === 'error') {
-      onError(new errors.StreamErrorEvent(event))
-    } else {
-      self._onEvent(event)
-    }
+    var events = json.parseJSONStreaming(data)
+
+    events.forEach(function(event) {
```
erickpintor, line 161: Is `forEach` ES5 compatible?

test/_json.test.js
```diff
@@ -0,0 +1,11 @@
+const { parseJSONStreaming } = require('../src/_json')
+
+describe('parseJSONStreaming', () => {
+  test('mutiple JSON objects in a message', () => {
+    // const jsonText = `{\"event\":\"version\",\"txnTS\":1603455982860000,\"data\":{\"ref\":{\"@ref\":{\"id\":\"278763355019149825\",\"collection\":{\"@ref\":{\"id\":\"Status\",\"collection\":{\"@ref\":{\"id\":\"collections\"}}}}}},\"ts\":{\"@ts\":\"2020-10-23T12:26:22.860Z\"},\"action\":\"update\",\"new\":{\"data\":{\"game\":1,\"players\":3,\"scores\":[305,27,9]}}}}{\"event\":\"version\",\"txnTS\":1603455983170000,\"data\":{\"ref\":{\"@ref\":{\"id\":\"278763355019149825\",\"collection\":{\"@ref\":{\"id\":\"Status\",\"collection\":{\"@ref\":{\"id\":\"collections\"}}}}}},\"ts\":{\"@ts\":\"2020-10-23T12:26:23.170Z\"},\"action\":\"update\",\"new\":{\"data\":{\"game\":1,\"players\":3,\"scores\":[306,27,9]}}}}`
```
erickpintor, line 5: What is this comment for?

src/stream.js
BrunoQuaresma: yes, it is.

test/_json.test.js
BrunoQuaresma: Nice catch. Removing it.

BrunoQuaresma:
> t does seem like a firefox only issue. Would you mind adding some notes in the code for it?

I added a note for it on top of the `parseJSONStreaming()` function.

```
/**
 * This function allow us to parse responses as streaming
 * currently, this is not allowed by Firefox only
 * https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API
 */
```

ryancharris (changes requested):
CI is failing

erickpintor:
Are those test failures legit or are we using the wrong build in the CI? Otherwise this change looks good to me.

BrunoQuaresma:
@erickpintor on local, this error is failing but it is also failing on master: 'wraps error events'. Could u check if this is working on your local?

BrunoQuaresma:
Since the error is also happening at master I'm merging this but I'm also investigating what is happening.

PR #360, opened 2020-10-28
Use http2 for streaming when is nodeJS

BrunoQuaresma:


ryancharris (commented):
Code is looking good 👍 Stinks we can't use ES6 syntax 😢

ryancharris (approved):

ryancharris:
@BrunoQuaresma Heads up that this is failing CI.

src/_http2.js
```diff
@@ -0,0 +1,39 @@
...
+  scheme = 'https',
+} = {}) {
+  var method = http2.constants.HTTP2_METHOD_POST
+  var client = http2.connect(url)
+
+  client.on('error', onError)
+
+  var request = client.request(
+    {
+      ...headers,
```
erickpintor, line 18: Is this ES5 compatible?

BrunoQuaresma: Good catch, I think this is not.

fauna-brecht: atm indeed "Uncaught ReferenceError: http2 is not defined"

PR #372, opened 2020-12-15
Fix for stringification to FQL which inserted nulls for zero-argument AND expose toFQL

fauna-brecht:
### Notes
This is related to this ticket:
[Jira Ticket](https://faunadb.atlassian.net/jira/software/c/projects/FE/issues/FE-642)

### Disclaimers
The code changes makes it possible for me to continue working, I'm currently working with a fork of this driver, do please check whether it doesn't break something else since I do not have that background on how the driver works. The PR is mainly here to  show where the problem is for the related ticket.
 
I have also exposed a function toFQL which I need for tooling. Feel free to separate both issues and/or rename it, doing a separate pull request is probably overkill.
If you choose not to make something like toFQL public, let me know, I'll have to include a chunk of code from the driver then in my tool.

### DIFF
--- a/src/Expr.js
+++ b/src/Expr.js
@@ -19,6 +19,10 @@ Expr.prototype.toJSON = function() {
   return this.raw
 }
 
+Expr.prototype.toFQL = function() {
+  return exprToString(this.raw)
+}
+
 var varArgsFunctions = [
   'Do',
   'Call',
--- a/src/Expr.js
+++ b/src/Expr.js
@@ -275,16 +275,13 @@ var exprToString = function(expr, caller) {
   var fn = keys[0]
   fn = convertToCamelCase(fn)
 
-  var args = []
-
-  keys.forEach(function(k) {
-    var v = expr[k]
-    if (v !== null) {
-      args.push(exprToString(v, fn))
-    }
-  })
-
-  args = args.join(', ')
+  // The filter prevents zero arity functions from having a null argument
+  // This only works under the assumptions
+  // that there are no functions where a single 'null' argument makes sense.
+  var args = keys
+    .filter(k => expr[k] !== null || keys.length > 1)
+    .map(k => exprToString(expr[k], fn))
+    .join(', ')
 
   return fn + '(' + args + ')'
 }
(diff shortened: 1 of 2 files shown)


marrony (changes requested):
Can you add tests for this, you can create a query like `new Query(x => Now())` and render it.

You can see examples of this on https://github.com/fauna/faunadb-js/blob/master/test/values.test.js#L395

fauna-brecht:
Thanks Marrony, I'll take a look once I find a snippet of time. 
Might take a while since I'm out for two weeks soon.. but who knows, I might feel an itch to code during my holidays

ryancharris (approved):

src/Expr.js
```diff
@@ -271,9 +275,13 @@ var exprToString = function(expr, caller) {
   var fn = keys[0]
   fn = convertToCamelCase(fn)
 
-  var args = keys.map(function(k) {
+  var args = []
+
+  keys.forEach(function(k) {
```
parkhomenko, line 280: I would probably, do it as:
`
keys
  .filter(function(k) { return expr[k] !== null })
  .map(function(k) { return exprToString(v, fn) })
`
but it can be a matter of preference anyway.

parkhomenko (dismissed):

src/Expr.js
fauna-brecht: Up to you, your approach is indeed cleaner and readability is important :)

marrony (changes requested):
The code change looks good, however the test coverage is very shy, for a bug like that we would need to add test for all 0-arity functions

fauna-brecht:
I changed the code (and added more tests while I was at it).
The code I adapted broke another case, specifically functions with other numbers of arguments than 0 that can take null arguments. Such as Select(...,..., null). 

These were not covered in tests so I added them.
The solution is not perfect, I'm merely providing **a** solution to help drive this forward and test my own tool that depends on this. If there is a way to determine whether a function is a 0-arity function, please replace the current filter statement. If not, we might want to opt to create special cases for those ?

parkhomenko (approved):

marrony (approved):

PR #373, opened 2020-12-16
ENG-2431: Switch from contatenated to line-separated JSON

erickpintor:
### Notes
To support for large JSON documents, we're moving from concatenated JSON objects to line-separated JSON streaming. This fix is intended to be backwards compatible while supporting the new line-separated format that is coming to cloud soon.

### How to test

With a recent nightly build: 

* Create a document;
* Open a stream for it;
* Issue an updated with a relatively large JSON document (about 4KB or more).

Without this fix, the JSON parsing on the received version should break. With this path, you shall receive the updated version normally as part of the document stream.

We should release this fix once ENG-2431 is in cloud.

### DIFF
--- a/src/_json.js
+++ b/src/_json.js
@@ -16,39 +16,43 @@ function parseJSON(json) {
   return JSON.parse(json, json_parse)
 }
 
-/**
- * This function allow us to parse responses as streaming
- * currently, this is not allowed by Firefox only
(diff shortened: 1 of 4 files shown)


rodrimaia (approved):
lgtm

szegedi (approved):
🐑

PR #380, opened 2021-01-15
Fix FaunaDate test to ignore timezone

fireridlle:
### Notes
Test "Values › date" doesn't respect timezone

### Actual result
```
Values › date

    expect(received).toEqual(expected) // deep equality

    Expected: "{\"@date\":\"1970-01-01\"}"
    Received: "{\"@date\":\"1969-12-31\"}"

      86 |     var test_date = new FaunaDate(new Date(1970, 0, 1))
      87 |     var test_date_json = '{"@date":"1970-01-01"}'
    > 88 |     expect(json.toJSON(test_date)).toEqual(test_date_json)
         |                                    ^
      89 |     expect(json.parseJSON(test_date_json)).toEqual(test_date)
      90 |   })
      91 | 

      at Object.test (test/values.test.js:88:36)
```

### Expected result
Test should be passed

### How to test
Set custom timezone (in my case GMT+3) and run test
```
jest --env=node --verbose=false ./test/values.test.js 
```

### DIFF
--- a/test/values.test.js
+++ b/test/values.test.js
@@ -83,7 +83,7 @@ describe('Values', () => {
   })
 
   test('date', () => {
-    var test_date = new FaunaDate(new Date(1970, 0, 1))
+    var test_date = new FaunaDate(new Date(Date.UTC(1970, 0, 1)))
     var test_date_json = '{"@date":"1970-01-01"}'
     expect(json.toJSON(test_date)).toEqual(test_date_json)
     expect(json.parseJSON(test_date_json)).toEqual(test_date)


erickpintor:
I'm surprised to see this test failing now since we've had folks in different timezones workin on the JS driver before. Do we know which part of the driver is rendering the wrong timezone? I don't oppose the fix. I'm just wondering what's the root cause.

fireridlle:
@erickpintor here is where FaunaTime transforms Date instance to iso string. so here is what i get
```javascript
new Date(1970, 0, 1).toISOString()
"1969-12-31T21:00:00.000Z"
```

erickpintor (approved):
I see now. It generates the string and truncates the time component. So, depending on your timezone *and* current time, you may fall back. I guess we were just lucky so far.

fireridlle:
closed as PR https://github.com/fauna/faunadb-js/pull/372 has this fix

PR #381, opened 2021-01-15
DRV-376 Fix request timeout

fireridlle:
### Notes
[Jira Ticket](https://faunadb.atlassian.net/browse/DRV-376)
Current driver pass a 60 seconds timeout down to the `cross-fetch` polyfill. However, neither `whatwg-fetch` of `node-fetch` are aware of it. They both support abort signals, though: https://github.com/node-fetch/node-fetch#request-cancellation-with-abortsignal.

### Screenshots
image

### DIFF
--- a/src/_http.js
+++ b/src/_http.js
@@ -4,6 +4,7 @@ var APIVersion = '4'
 
 var parse = require('url-parse')
 var util = require('./_util')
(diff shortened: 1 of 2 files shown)


erickpintor (approved):
LGTM

parkhomenko (approved):

PR #385, opened 2021-01-25
DRV-337 driver does not graciously fail when we go over the limit

fireridlle:
### Notes
[driver does not graciously fail when we go over the limit](https://faunadb.atlassian.net/browse/DRV-337)
JS driver does not return errors when going over the stream limit.

Unfortunately, there is no evidence from the backend that we can relly and recognize that client faced to limit. backend just keep a connection in a pending state.

Currently, the issue happens only for the JS browser driver, however, as soon as http2 would be implemented for nodejs, we would be faced the same issue for nodejs as well. because of http2 multiplexing feature, the number of concurrent requests per domain is not limited to 6-8, but it is virtually unlimited. It is virtually unlimited in the sense that browsers and servers may limit the number of concurrent requests via the configuration parameter called SETTINGS_MAX_CONCURRENT_STREAMS. if it’s client settings, this means how many concurrent streams the client can handle it’s backend settings, this means how many concurrent streams the backend can accept. i haven’t found any change to increase this param as it’s specified in the source code of browsers. however, i found that the backend has the following settings MAX_CONCURRENT_STREAMS = 100, probably we can increase it and check if requests are stuck or not. 

Anyway, JS driver has to throw an error for such a case, the workaround would be using the same approach with the abort controller as for queries. basically, it aborts request by timeout. 

### How to test
Sample code
```javascript
  secret: 'your secret',
  timeout: 5
})

function startStream() {
  stream = client.stream.document(q.Ref(q.Collection('users'),'282652881944838658'))
  .on('snapshot', snapshot => {
    console.log('snapshot')
    startStream()
  })
  .on('version', version => {
    console.log('version')
  })
  .on('error', error => console.error('Error: %s', error))
  .start()
}

startStream()
```
The expected result would be an error when the number of open streams rich limit. Also good to test that 1 open stream is not aborted by timeout

### DIFF
--- a/src/_http.js
+++ b/src/_http.js
@@ -64,7 +64,7 @@ HttpClient.prototype.syncLastTxnTime = function(time) {
  * @param {?string} body The HTTP request body.
  * @param {?string} query The HTTP request query parameters.
  * @param {?Object} options The request options.
- * @param {?Object} options.signal An abort signal object.
+ * @param {?Object} options.abortController An abort controller object.
  * @param {?Object} options.fetch A Fetch API compatible function.
  * @param {?string} options.secret A FaunaDB secret.
  * @param {?string} options.queryTimeout A FaunaDB query timeout.
@@ -77,7 +77,7 @@ HttpClient.prototype.execute = function(method, path, body, query, options) {
   url.set('query', query)
   options = util.defaults(options, {})
 
-  var signal = options.signal
+  var abortController = options.abortController || new AbortController()
   var fetch = options.fetch || this._fetch
   var secret = options.secret || this._secret
   var queryTimeout = options.queryTimeout || this._queryTimeout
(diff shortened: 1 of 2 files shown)


src/_http.js
```diff
@@ -90,16 +90,14 @@ HttpClient.prototype.execute = function(method, path, body, query, options) {
   headers['X-Query-Timeout'] = queryTimeout
 
   var timeout
-  if (!signal && this._timeout) {
-    var abortController = new AbortController()
-    signal = abortController.signal
-    timeout = setTimeout(abortController.abort, this._timeout)
+  if (this._timeout) {
```
erickpintor, line 93: I _think_ this will apply the same default query timeout to both queries and streams, right? If so, that's undesirable as streams are suppose to be long running handlers that don't timeout.

erickpintor (changes requested):

src/_http.js
fireridlle: not really. it's a timeout request to api POST /stream, so if this api doesn't respond in timeout, it will cancel the request. but as soon as request success, driver create stream

erickpintor (approved):
🐑 it!

n400:
We probably don't need this anymore since we switched to HTTP/2, right? In any case, it will need some changes, so closing it

PR #389, opened 2021-01-26
chore: Added tests to reflect the error event delegation

parkhomenko:
### Notes
[Jira Ticket](https://faunadb.atlassian.net/browse/DRV-444)

### How to test
This fix doesn't introduce any changes to the client code, but rather adds missing tests to reflect that if there is an unexpected error in opened stream connection happened, we have 2 errors sent to the client: from onError and onEnd events.

### DIFF
--- a/test/stream.test.js
+++ b/test/stream.test.js
@@ -155,8 +155,8 @@ describe('StreamAPI', () => {
 
     test('wraps delegated error events if stream closed unexpectedly', async done => {
       testErrorEvent(done, error => {
-        if (error.code === undefined) {
-          expect(error.name).toEqual('TypeError')
+        if (error instanceof TypeError) {
+          expect(error.message).toEqual('network error')
         }
       })
     })
--- a/test/stream.test.js
+++ b/test/stream.test.js
@@ -113,7 +113,7 @@ describe('StreamAPI', () => {
       expect(() => stream.start()).toThrow(Error)
     })
 
-    test('wraps error events', async done => {
+    async function testErrorEvent(done, checkErrorCallback) {
       let role = await client.query(
         q.CreateRole({
           name: util.randomString('role'),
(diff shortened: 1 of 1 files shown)


test/stream.test.js
```diff
@@ -136,13 +136,29 @@ describe('StreamAPI', () => {
...
-          done()
-        })
-        .start()
+        }
+      })
+    })
+
+    test('wraps delegated error events if stream closed unexpectedly', async done => {
+      testErrorEvent(done, error => {
+        if (error.code === undefined) {
```
marrony, line 158: what happens if `error.code !== undefined`? what should be the assertion in this case?

parkhomenko: i wanted to test the error event that happens when we destroy the secret while the stream connection is already opened:
https://github.com/parkhomenko/faunadb-js/blob/drv-444/test/stream.test.js#L129-L142

the issue was that in that case, the ".error" event triggered 2 times:
- the error object is sent like this: {code: 'permission denied', description: 'Authorization lost during stream evaluation.'}
- the internal `onEnd` event is triggered and delegates to the 'onError' event: https://github.com/parkhomenko/faunadb-js/blob/drv-444/src/stream.js#L187-L193, and we have a different `error` object, that doesn't have fields `code` and `description` in it

the first test checks the first case: https://github.com/parkhomenko/faunadb-js/blob/drv-444/test/stream.test.js#L145-L154 (`in this case error.code !== undefined`)
the second test checks that we got the `error` event delegated from `onEnd` event, and `error.code` is undefined in this scenario

so if `error !== undefined` the assertion defined here: https://github.com/parkhomenko/faunadb-js/blob/drv-444/test/stream.test.js#L148-L151 checks the `error` object

marrony: I understand your arguments, but the conditional assertion on tests looks wrong to me. I would like to know others opinion regarding this.

parkhomenko: could you advise who from the team can I add to this PR to take a look at it?

marrony: @erickpintor @evbruno @rodrimaia can you take a look on this?

erickpintor: I'm okay with the error type. I think the assertion could be a little bit better.

```
if (error instanceof TypeError) {
  expect(error.message).toEqual('network error')
}
```

parkhomenko: Good point. I've changed it as you suggested. Thanks!

erickpintor (approved):
🐑 it.

rodrimaia (approved):

PR #395, opened 2021-02-01
Decrease bundle size

fireridlle:
don't inject http,https,os,utils polyfills. they are required only inside condition isNodeEnv, so we are safe to remove them from the browser bundle

extract from the heavy url-parse package only what we need

Before
image

After
image

Note: screenshots above were made from non minified bundle. minified bundle about 78KB.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10868,11 +10868,6 @@
       "integrity": "sha1-nsYfeQSYdXB9aUFFlv2Qek1xHnM=",
       "dev": true
     },
-    "querystringify": {
-      "version": "2.1.1",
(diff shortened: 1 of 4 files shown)


parkhomenko (approved):

erickpintor (approved):
LGTM

fireridlle:
@n400 there are some more improvements and investigations we can do. for example, replace `fn-annotate` package with our own implementation will save about 0.6kb (for minified version) as well as review `query.js` which currently takes 41% of bundle size.

one thing more, we probably could extract streaming feature to separate package, so users who don't need this feature will install only fauna core features

PR #401, opened 2021-02-04
Revert abort controller and more sophisticated nodejs env detection

fireridlle:
Abort controller polyfill has been changed because of cloudflare issue
```
(node:16784) UnhandledPromiseRejectionWarning: TypeError: AbortController is not a constructor
    at r.execute (/Users/szinkevych/projects/faunadb/fauna-express/functions/index.js:164:2194)
    at d._execute (/Users/szinkevych/projects/faunadb/fauna-express/functions/index.js:325:32329)
    at d.query (/Users/szinkevych/projects/faunadb/fauna-express/functions/index.js:325:31673)
```
However, that polyfill simply rejects a promise but not terminating a request. So we have to return the previous AbortController. Cloudflare team doesn't intend to implement AbortController https://community.cloudflare.com/t/timeout-with-fetch/25249/4. The possible workaround has been added to Readme.md

One more issue appears at Cloudflare, as wrangler build using NodeJS env but Cloudflare worker running as service worker (at browser) `isNodeEnv` doesn't behave correctly, therefore it leads to issue as we remove `http`,`https` polyfills from browser bundle to minify size.

### DIFF
--- a/README.md
+++ b/README.md
@@ -246,11 +246,11 @@ const client = new faunadb.Client({
 
 ### Using with Cloudflare Workers
 
-Cloudflare Workers has neither XMLHttpRequest nor fetch in global scope.
-Therefore, cross-fetch package is unable to inject fetch, and throws an error.
-fetch function is injected via closure, so the workaround would be to pass
-the fetch objects when initiating the FaunaDB client config. Cloudflare also
-doesn't support AbortController, which terminates requests as well as streams.
+Cloudflare Workers have neither XMLHttpRequest nor fetch in the global scope.
+Therefore, the `cross-fetch` package is unable to inject its own `fetch()` function, and throws an error.
+The `fetch()` function is injected via a closure, so the workaround would be to pass
+the fetch objects when initiating the FaunaDB client config. Cloudflare Workers also
+doesn't support the use of an AbortController, which terminates requests as well as streams.
 Here is a workaround:
 
 ```javascript
--- a/README.md
+++ b/README.md
@@ -242,6 +242,31 @@ const client = new faunadb.Client({
 })
 ```
 
+## Knowing issues
+
+### Using at Cloudflare workers
+
+Cloudflare workers doesn't have XMLHttpRequest neither fetch in global scope,
+therefore cross-fetch package unable to inject fetch and throw an error.
+fetch function is injected via closure, so workaround would be to pass
+those fetch object to initiate FaunaDB client config. Cloudflare also doesn't support AbortController which terminate requests as well as streams. Here is workaround
+
+```javascript
+const c = new faunadb.Client({
+  secret: 'your secret',
+  fetch: (url, params) => {
+    const signal = params.signal
+    delete params.signal
+    const abortPromise = new Promise(resolve => {
+      if (signal) {
+        signal.onabort = resolve
+      }
+    })
+    return Promise.race([abortPromise, fetch(url, params)])
+  },
+})
+```
+
 ## Client Development
 
 Run `yarn` to install dependencies.
--- a/README.md
+++ b/README.md
@@ -242,14 +242,16 @@ const client = new faunadb.Client({
 })
 ```
 
-## Knowing issues
+## Known issues
 
-### Using at Cloudflare workers
+### Using with Cloudflare Workers
 
-Cloudflare workers doesn't have XMLHttpRequest neither fetch in global scope,
-therefore cross-fetch package unable to inject fetch and throw an error.
-fetch function is injected via closure, so workaround would be to pass
-those fetch object to initiate FaunaDB client config. Cloudflare also doesn't support AbortController which terminate requests as well as streams. Here is workaround
+Cloudflare Workers has neither XMLHttpRequest nor fetch in global scope.
+Therefore, cross-fetch package is unable to inject fetch, and throws an error.
+fetch function is injected via closure, so the workaround would be to pass
+the fetch objects when initiating the FaunaDB client config. Cloudflare also
+doesn't support AbortController, which terminates requests as well as streams.
+Here is a workaround:
 
 ```javascript
 const c = new faunadb.Client({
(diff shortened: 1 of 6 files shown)


tigger9flow (dismissed):

README.md
```diff
@@ -242,6 +242,31 @@ const client = new faunadb.Client({
 })
 ```
 
+## Knowing issues
```
n400, line 245: ## Known issues

### Using with Cloudflare Workers

Cloudflare Workers has neither XMLHttpRequest nor fetch in global scope.
Therefore, cross-fetch package is unable to inject fetch, and throws an error.
fetch function is injected via closure, so the workaround would be to pass
the fetch objects when initiating the FaunaDB client config. Cloudflare also 
doesn't support AbortController, which terminates requests as well as streams. 
Here is a workaround:

n400: just some minor grammatical edits

n400 (changes requested):

README.md
fireridlle: done

README.md
```diff
@@ -242,6 +242,33 @@ const client = new faunadb.Client({
 })
 ```
 
+## Known issues
+
+### Using with Cloudflare Workers
+
+Cloudflare Workers has neither XMLHttpRequest nor fetch in global scope.
```
faunaee, line 249:
```suggestion
Cloudflare Workers have neither XMLHttpRequest nor fetch in the global scope.
```

README.md
```diff
@@ -242,6 +242,33 @@ const client = new faunadb.Client({
...
+Therefore, cross-fetch package is unable to inject fetch, and throws an error.
```
faunaee, line 250:
```suggestion
Therefore, the `cross-fetch` package is unable to inject its own `fetch()` function, and throws an error.
```

README.md
```diff
@@ -242,6 +242,33 @@ const client = new faunadb.Client({
...
+fetch function is injected via closure, so the workaround would be to pass
```
faunaee, line 251:
```suggestion
The `fetch()` function is injected via a closure, so the workaround would be to pass
```

README.md
```diff
@@ -242,6 +242,33 @@ const client = new faunadb.Client({
...
+the fetch objects when initiating the FaunaDB client config. Cloudflare also
```
faunaee, line 252: Objects? Do you mean "function"? If not, it would be good to provide a bit more context here.

```suggestion
the fetch objects when initiating the FaunaDB client config. Cloudflare Workers also
```

README.md
```diff
@@ -242,6 +242,33 @@ const client = new faunadb.Client({
...
+doesn't support AbortController, which terminates requests as well as streams.
```
faunaee, line 253:
```suggestion
doesn't support the use of an AbortController, which terminates requests as well as streams.
```
When you say "which terminates", the subject is ambiguous. Does an AbortController terminate requests/streams, or does the lack of an AbortController cause that?

README.md
```diff
@@ -242,6 +242,33 @@ const client = new faunadb.Client({
...
+    const signal = params.signal
+    delete params.signal
+    const abortPromise = new Promise(resolve => {
+      if (signal) {
+        signal.onabort = resolve
+      }
+    })
+    return Promise.race([abortPromise, fetch(url, params)])
+  },
+})
```
faunaee, line 269: Nice! I haven't seen that done before.

faunaee (changes requested):
I've only reviewed the updates to the README file.

I noted a few grammatical fixes, and provided some suggested edits to make the additions clearer. I asked a couple of questions where the answers should help to further clarify the content.

Aside from those, looks good!

README.md
fireridlle: yes, it's about function

README.md
fireridlle: AbortController terminates requests. Cloudflare doesn't have AbortController. so even if a user decides to close a stream, it would be still alive

tigger9flow (approved):

README.md
```diff
@@ -242,6 +242,33 @@ const client = new faunadb.Client({
...
+  secret: 'your secret',
+  fetch: (url, params) => {
+    const signal = params.signal
+    delete params.signal
+    const abortPromise = new Promise(resolve => {
+      if (signal) {
+        signal.onabort = resolve
+      }
+    })
+    return Promise.race([abortPromise, fetch(url, params)])
```
marrony, line 267: Where this `fetch` come from? Shouldn't we include the require statements in the snippet?

README.md
```diff
@@ -242,6 +242,33 @@ const client = new faunadb.Client({
...
+the fetch objects when initiating the FaunaDB client config. Cloudflare Workers also
+doesn't support the use of an AbortController, which terminates requests as well as streams.
+Here is a workaround:
+
+```javascript
+const c = new faunadb.Client({
+  secret: 'your secret',
+  fetch: (url, params) => {
+    const signal = params.signal
+    delete params.signal
```
marrony, line 261: why `delete` is necessary here?

README.md
fireridlle: no, Cloudflare inject it into a function handler

README.md
fireridlle: Cloudflare doesn't implement signal into their fetch, if the signal property is not removed, an error would be thrown

marrony (approved):

n400 (approved):

parkhomenko (approved):

README.md
faunaee: That means that `objects` needs to be replaced with `function`.

README.md
faunaee: Thanks for the information. I'm still not completely certain that I know how to write the last sentence to resolve the ambiguity; let me know if this sounds correct:

`... the use of an AbortController, which can cause request and stream termination without sufficient error handling.`

Issue #289, opened 2020-06-08
Observer option passed to query is not handled


steida:
This:

```js
if (self._observer != null) {
  self._observer(requestResult, self)
}
```

My use case is to create a default Client instance and pass the secret to the query method because of a different context. If I will add observer in the constructor, TypeScript type requires also a secret, which I don't have yet. 

A workaround is possible, but it would be better if the observer from query options would work.

```
var observer = self._observer || (options && options.observer)

if (observer != null) {
  observer(requestResult, self)
}
```

BrunoQuaresma:
Sometime ago we had a discussion about this internally and I'll bring this up again. Thanks for let us know your use case.

fireridlle:
Hi @steida . would that works for you?
https://github.com/fauna/faunadb-js/pull/402

note: an observer that passed to the Client constructor would be overwritten by an observer passed to the query.

steida:
LGTM

PR #402, opened 2021-02-04
Accept observer for query

fireridlle:
### Notes
[DRV-437](Observer option passed to query is not handled)

issue: https://github.com/fauna/faunadb-js/issues/289

### DIFF
--- a/test/client.test.js
+++ b/test/client.test.js
@@ -82,6 +82,18 @@ describe('Client', () => {
     )
   })
 
+  test('query observer', async () => {
+    var clientObserver = jest.fn()
+    var queryObserver = jest.fn()
+    var client = util.getClient({ observer: clientObserver })
+
+    await client.query(query.Now(), {
+      observer: queryObserver,
+    })
+    expect(clientObserver).toBeCalledTimes(0)
+    expect(queryObserver).toBeCalledTimes(1)
+  })
+
   test('keeps connection alive', () => {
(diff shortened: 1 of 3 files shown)


parkhomenko (dismissed):

test/client.test.js
```diff
@@ -82,6 +82,18 @@ describe('Client', () => {
...
 
+  test('query observer', async () => {
+    var clientObserver = jest.fn()
+    var queryObserver = jest.fn()
+    var client = util.getClient({ observer: clientObserver })
+
+    await client.query(query.Now(), {
+      observer: queryObserver,
+    })
+    expect(clientObserver).toBeCalledTimes(0)
```
marrony, line 93: Shouldn't we call both observers here? I imagine someone doing some kinda of global metrics aggregation using the `clientObserver` and a per-query metrics using `queryObserver`, as is, if we do not call the `clientObserver` the customer would need to do the global aggregation in the local too.

fireridlle: on the one hand, that makes sense. but on the other hand, all options passed to query override values passed during instantiate the client object.

anyway, i commit the requested changes. let me know if it's ok, please

parkhomenko (approved):

test/client.test.js
fireridlle: @marrony what do you think?

erickpintor (approved):
lgtm

Issue #384, opened 2021-01-25
Missing type definitions for RequestResult


michaelmerrill:
The property definitions for [RequestResult](https://github.com/fauna/faunadb-js/blob/master/src/types/RequestResult.d.ts) are not fully defined. For example, the following code must be ignored due to `responseContent` being defined as `object`.

```(typescript)
const errors = error.requestResult.responseContent.errors;  // Property 'errors' does not exist on type 'object'.
```

parkhomenko:
@michaelmerrill 
Your issue should be fixed in the latest version of the driver.

Issue #341, opened 2020-09-19
TypeScript error


rishabhpoddar:
This is specifically for using this package in NodeJS.

Version of package: `^3.0.1`

NodeJS version: `v13.11.0`

Operating System: MacOS 10.15.6

Version of TypeScript: `3.8.3`

Error:
```
../node_modules/faunadb/src/types/Client.d.ts:15:18 - error TS2304: Cannot find name 'fetch'.

15   fetch?: typeof fetch
```

dwlizlo:
As a workaround, you can use the solution below. It will lie to the compiler in a sense, so remember to remove it once the issue is fixed.

```
"compilerOptions": {
        "lib": [ "dom" ],
   }
```

parkhomenko:
@rishabhpoddar @dwlizlo 
Your issue should be fixed in the latest version of the driver.

PR #405, opened 2021-02-08
Add missing type definitions

fireridlle:
### Notes
[DRV-435](https://faunadb.atlassian.net/browse/DRV-435)

Linked issues
https://github.com/fauna/faunadb-js/issues/384
https://github.com/fauna/faunadb-js/issues/341

### How to test
tsconfig.json should have "lib": ['es2019']. in that case ts will not inject type for `fetch` (will only if lib has `DOM`)
No typescript error should be thrown
```javascript

import { Client, errors, FaunaHttpErrorResponseContent, query, values } from './index';

const client = new Client({
    secret: '',
    observer: (res => {
        if((res.responseContent as FaunaHttpErrorResponseContent).errors) {
            console.error('observer receive error')
            console.error(res.responseContent)
        }
    })
})


client.query<values.FaunaTime>(query.Now())
.then(now => {
    console.info('now ', now.date)
})
.catch((err: errors.FaunaHTTPError) => {
    console.error('query catch error')
    console.error(err.requestResult.responseContent.errors)
})

```

### DIFF
--- a/src/RequestResult.js
+++ b/src/RequestResult.js
@@ -94,19 +94,3 @@ Object.defineProperty(RequestResult.prototype, 'timeTaken', {
 })
 
 module.exports = RequestResult
-
-// /**
-//  * A structure containing the response errors for a given FaunaDB request.
-//  * Provided to an observer function optionally defined in the {@link Client} constructor.
-//  *
-//  * @param {'GET' | 'POST' | 'PUT' | 'PATCH' | 'DELETE'} method
-//  *   The HTTP method used in the request.
-//  * @param {Array.<{code: string, description: string}>} errors
-//  * @constructor
-//  */
-// function FaunaHttpErrorResponseContent(errors) {
-//   /** @type {Array.<{code: string, description: string}>} */
-//   this.errors = errors
-// }
-
-// module.exports.FaunaHttpErrorResponseContent = FaunaHttpErrorResponseContent
--- a/src/RequestResult.js
+++ b/src/RequestResult.js
@@ -16,7 +16,7 @@
  *   The request data.
  * @param {string} responseRaw
  *   The unparsed response data, as a string.
- * @param {object} responseContent
+ * @param {object | FaunaHttpErrorResponseContent} responseContent
  *   The response data parsed as JSON.
  * @param {number} statusCode
  *   The HTTP response status code.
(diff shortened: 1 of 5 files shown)


parkhomenko (approved):

erickpintor (dismissed):
Neat! 👍

src/types/Client.d.ts
```diff
@@ -26,7 +28,7 @@ export interface QueryOptions {
 
 export default class Client {
   constructor(opts?: ClientConfig)
-  query<T = object>(expr: ExprArg, options?: QueryOptions): Promise<T>
+  query<T extends object = object>(expr: ExprArg, options?: QueryOptions): Promise<T>
```
tigger9flow, line 31: Fauna query result is not limited to type `object` only, e.g. for the query below it returns just a plain number:
```
.query(Q.Add(2, 2)) // Promise<number>
```
Another thing is it'd be great to use the `unknown` type as a default for any generic type in Fauna. But not sure whether it's backward-compatible :thinking_face:

src/RequestResult.js
```diff
@@ -94,3 +94,19 @@ Object.defineProperty(RequestResult.prototype, 'timeTaken', {
 })
 
 module.exports = RequestResult
+
```
tigger9flow, line 97: Is it redundant?

src/types/Client.d.ts
```diff
@@ -13,7 +15,7 @@ export interface ClientConfig {
   port?: number
   timeout?: number
   queryTimeout?: number
-  observer?: (res: RequestResult, client: Client) => void
+  observer?: <T extends object = object>(res: RequestResult<T | errors.FaunaHTTPError>, client: Client) => void
```
tigger9flow, line 18: Same for `object` type restriction as mentioned below

fireridlle: not really. observer accepts the full response or error

src/types/Client.d.ts
fireridlle: removed `extends`. not sure about the `unknown` type. probably we will think about that no sooner than migrate driver code to typescript. so we can use that types during development and check how things work

src/RequestResult.js
fireridlle: yes

parkhomenko (approved):

tigger9flow (approved):

PR #413, opened 2021-02-18
DRV-367: add api-version for Query

fireridlle:
### Notes
[DRV-367 API version should always be reported when the current version is not 3 ](https://faunadb.atlassian.net/browse/DRV-367)

### How to test
```
  client.query(
    q.Query(q.Lambda('x', q.Var('x')))
  )
  .then((ret) => console.log(ret))
  .catch((err) => console.error('Error: %s', err))
```

### DIFF
--- a/src/values.js
+++ b/src/values.js
@@ -361,7 +361,7 @@ util.inherits(Query, Value)
 
 wrapToString(Query, function() {
   const metadata = {
-    'api-version': +this.value.api_version,
+    api_version: +this.value.api_version,
   }
   return (
     'Query(' + Expr.toString(this.value) + ', ' + JSON.stringify(metadata) + ')'
--- a/src/values.js
+++ b/src/values.js
@@ -363,9 +363,23 @@ wrapToString(Query, function() {
   const metadata = {
     api_version: +this.value.api_version,
   }
-  return (
-    'Query(' + Expr.toString(this.value) + ', ' + JSON.stringify(metadata) + ')'
-  )
+
+  const notEmptyMetadata = Object.keys(metadata).some(key => metadata[key])
+
+  let stringified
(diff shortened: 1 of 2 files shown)


marrony:
Api version should not be included in this case because `Query` function doesn’t accept it, users are not allowed to set the api version on query types

faunaee:
@marrony I'm likely misreading the source, but how does `wrapToString` affect the parameters sent to `Query`?

fireridlle:
> Api version should not be included in this case because `Query` function doesn’t accept it, users are not allowed to set the api version on query types

probably i don't get the statement, but it's only string representation, whenever `Query` object should be represented as a string (for example print at the console).

parkhomenko (approved):

src/values.js
```diff
@@ -360,7 +360,26 @@ function Query(value) {
...
+      Expr.toString(this.value),
+      ', ',
+      JSON.stringify(metadata),
+      ')',
+    ]
+  } else {
+    stringified = ['Query(', Expr.toString(this.value), ')']
+  }
+
+  return stringified.join('')
```
evbruno, line 382:
```suggestion
  const notEmptyMetadata = Object.keys(metadata).some(key => metadata[key])

   if (Object.keys(metadata).some(key => metadata[key])) {
     return [
       'Query(',
       Expr.toString(this.value),
       ', ',
       JSON.stringify(metadata),
       ')',
     ].join('')
   }

   return ['Query(', Expr.toString(this.value), ')'].join('')
```

I usually don't like multiple `returns, but I think it would be fine here.

Anyway, the PR is good and feel free to discard this suggestion :)

evbruno (approved):
one minor comment, but this LGTM

PR #414, opened 2021-02-19
Fix merging headers

fireridlle:
### Notes
[DRV-505 faunadb-js 4.1.0 release broke downstream products](https://faunadb.atlassian.net/browse/DRV-505)

### DIFF
--- a/src/_http/index.js
+++ b/src/_http/index.js
@@ -30,7 +30,7 @@ function HttpClient(options) {
       })
   this._baseUrl = options.scheme + '://' + options.domain + ':' + options.port
   this._secret = options.secret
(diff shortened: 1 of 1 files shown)


parkhomenko (approved):
Thanks for the quick fix!

parkhomenko:
Does it make sense to add a unit test as well to check this issue that happened in fauna-shell won't be the cause in the future?

fireridlle:
> Does it make sense to add a unit test as well to check this issue that happened in fauna-shell won't be the cause in the future?

I think this happens just because of not well familiarity with the existing code base. So internal function is not working as i expected

tigger9flow (approved):

PR #425, opened 2021-03-02
DRV-522 always pass query timeout

fireridlle:
### Notes
[DRV-522](https://faunadb.atlassian.net/browse/DRV-522)

### DIFF
--- a/src/_http/index.js
+++ b/src/_http/index.js
@@ -31,9 +31,9 @@ function HttpClient(options) {
   this._baseUrl = options.scheme + '://' + options.domain + ':' + options.port
   this._secret = options.secret
   this._headers = Object.assign({}, options.headers, getDefaultHeaders())
-  this._queryTimeout = options.queryTimeout
   this._lastSeen = null
   this._timeout = Math.floor(options.timeout * 1000)
+  this._queryTimeout = options.queryTimeout || this._timeout
 }
 
 /**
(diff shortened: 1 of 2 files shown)


parkhomenko (approved):

tigger9flow (approved):

erickpintor (changes requested):
We should unify request and query timeout. Having multiple dimensions can be confusing.

fireridlle:
> We should unify request and query timeout. Having multiple dimensions can be confusing.

for sure it would be better. but what if some users use timeout prop and some queryTimeout prop. so i decided to keep them both for backward compatibility

erickpintor:
There isn't a good reason to control them separately. If request timeout is shorter than query timeout, the client gives up. If the query timeout is shorter, then you get a server timeout. You inevitably get a timeout at the minimum value set. Besides, removing the optional parameter doesn't break anyones code, it just gets ignored. I think it's safe to drop it.

erickpintor (dismissed):
LGTM

tigger9flow (dismissed):

fireridlle:
@erickpintor @tigger9flow could you re-approve please. just fixed the merge conflict.

tigger9flow (approved):

erickpintor (approved):
🐑 it

PR #430, opened 2021-03-04
test: check auth0 setup correctly

fireridlle:
### Notes
[DRV-543](https://faunadb.atlassian.net/browse/DRV-543)

Auth0 integration test works. However, auth0 return an error with status code 200 which lead to the promise resolved successfully and we proceed. In that case, the test failed with a not really clear message

### DIFF
--- a/test/auth.test.js
+++ b/test/auth.test.js
@@ -104,6 +104,12 @@ describe('auth', () => {
       })
     })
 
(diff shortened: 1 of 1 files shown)


parkhomenko (approved):

erickpintor (approved):
LGTM

PR #433, opened 2021-03-05
fix(htt2Adapter): set http2 session idle time

fireridlle:
### Notes
[DRV-523](https://faunadb.atlassian.net/browse/DRV-523)

This is related to how http2 sessions work, a http2 session should be reused across multiple calls thus should have some lifetime specified. By default, it’s set to 60 secs which is reasonable for most web services (when having continuous incoming traffic). But unfortunately, I haven’t considered such cases like using fauna-shell - it definitely doesn’t make sense to keep this session for a long time in such scenarios.
The tricky question is to decide the reasonable/versatile lifetime for http2 sessions as the driver has a lot of use-cases in different environments. We could accept this http2 session lifetime via params when constructing the Client e.g. http2Lifetime but would it confuse the customer to configure it by themself?

Solution is taken from [google client](https://github.com/googleapis/nodejs-googleapis-common/blob/master/src/http2.ts#L208)

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "browserify": "browserify index.js --standalone faunadb -o dist/faunadb.js",
     "browserify-min": "browserify index.js --standalone faunadb | terser -c -m --keep-fnames --keep-classnames -o dist/faunadb-min.js",
     "prettify": "prettier --write \"{src,test}/**/*.{js,ts}\"",
-    "test": "jest --env=node --verbose=false ./test",
+    "test": "jest --env=node --verbose=false --forceExit ./test",
     "semantic-release": "semantic-release",
     "wp": "webpack"
   },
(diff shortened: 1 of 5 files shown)


tigger9flow (approved):

erickpintor:
Alternatively we could add a `close()` function to the deriver instead, thus letting customers release underlying held resources as the database client is no longer needed.

tigger9flow:
@erickpintor Sounds good for me, we can add it in a breaking change version and make `http2` session timeout more reasonable like it's was before (60secs). For now, we should stick with keeping API as it is and not require customers to call another method to make things work

erickpintor (approved):
This is fine for a patch version but, as mentioned, I believe a closing method to the client interface is desirable feature still.

n400:
We might need to revisit this. If it's been 60 secs in production for a while now, could changing this to 500ms introduce a breaking change to anyone's apps? Maybe we should just add the optional closing method as Erick suggested. Which is "less breaking"?

PR #441, opened 2021-03-16
feature: configurate building tool to work with es5+esm and es6+

fireridlle:
### Notes
[DRV-553 Configure the new bundling tool (webpack) to work with ES2015 module. Three types of package delivery should be configured - UMD, CJS, ESM.](https://faunadb.atlassian.net/browse/DRV-553)

To allow independent imports, we should use the ECMAScript modules (ES2015 modules) system. It works great with TypeScript as well, but ES2015 modules are not supported in many environments without bundling (webpack/rollup, etc). This way, the “faunadb” should be implemented as a multi-platform package exposing modules in ES2015 format as well as old-fashioned UMD and CJS module systems. Path to build (based on ES2015 modules) folder should be specified under 

Path to build specified inside package.json by using fields
- index: common js modules system + es5 code
- main: es2015 modules system + es5 code
- es2015: es6 code

New npm command:
- bundle:view: open dev server to analyze bundle size
- bundle:stats: generate stat.json with information about bundle size
- bundle:diff: compare two stat.json. mainly used for pull request to check size diff between head & base
- build:clean: remove build artifacts
- build:browser: bundle umd for browser (that what comes to cdn)
- build:cjs: transpile to es5 + common js modules
- build:esm5: transpile to es5 + esm 
- build:package: run all build:* command

NodeJS minimum support: lts/dubnium -> v10.24.0
cdn bundle support: last 2 versions
```
and_chr 88
and_ff 85
and_qq 10.4
and_uc 12.12
android 81
baidu 7.12
chrome 88
chrome 87
chrome 86
edge 88
edge 87
firefox 85
firefox 84
firefox 78
ie 11
ios_saf 14.0-14.4
ios_saf 13.4-13.7
ios_saf 12.2-12.4
kaios 2.5
op_mini all
op_mob 59
opera 73
opera 72
safari 14
safari 13.1
samsung 13.0
samsung 12.0
```

In addition, this pr has Github action that runs on each PR and compares bundle size from head and base. Note, by bundle size I mean cdn bundle. Tracking the real output for users is impossible as soon as we introduce esm, their bundle tools would be able to automatically thee-shaking and import only what they use rather than the whole package. Also, their browser support could be less, therefore, number of polifyls would be less as well.


parkhomenko (approved):

erickpintor (commented):
It looks good to me but I'm not the best person to approve it. I suspect @rodrimaia will have more insight.

PR #446, opened 2021-03-23
(WIP) Feature/drv 549 modularize

fireridlle:
### Notes
[Jira Ticket]()

### Screenshots


README.md
```diff
@@ -126,6 +126,32 @@ createP.then(function(response) {
...
 
+#### ES6 modules
+
+This is recommended way to inject faunadb-js. Many build tools supports three shaking which relies on the static structure of ES6 module syntax, i.e. `import` and `export`.
+
+```
+import { Client, query } from 'faunadb'
+
+const client =new Client({secret: 'your_key_here'})
+client.query(query.Now()).then(console.info)
```
faunaee, line 137:
```suggestion
import { Client, q } from 'faunadb'
const client = new Client({secret: 'your_secret_here'})
client.query(q.Now()).then(console.info)
```

All of our documented examples use `q` as the namespace for FQL functions. If we don't make that change, the `client.query(query.Now())` can confuse newbies.

README.md
```diff
@@ -126,6 +126,32 @@ createP.then(function(response) {
...
+```
+
+Another way to import queries functions
+
+```
+import { Now, Select, Var, FaunaObject, FaunaFunction } from 'faunadb/query'
+query.Object has been replaced to FaunaObject
+query.Function has been replaced to FaunaFunction
```
faunaee, line 145: Is this code? If not:

```suggestion
// query.Object has been imported as FaunaObject
// query.Function has been imported as FaunaFunction
```

README.md
```diff
@@ -126,6 +126,32 @@ createP.then(function(response) {
...
+```
+
```
faunaee, line 147: Here is a great place to remind users that `Function` and `Object` are reserved words in JavaScript, so if they want to import as non-namespaced names, they need to be imported with custom names, such as `FFunction` and `FObject`.

faunaee (commented):
Just a few suggestions.

With this PR, would `require('faunadb')` stop working at all? I'm wondering if this change would affect any existing client scripts.

README.md
fireridlle: we can update it it
```
import {query as q} from 'faunadb'
```
it's the equivalent to what we currently have
```
const faunadb = require('faunadb')
const q = faunadb.query;
```
but everyone familiar with both ecma script modules and commonjs modules

README.md
fireridlle: no, this is just notes that we have renamed those queries as `OBject` and `Function` are reserved world in js

Issue #447, opened 2021-03-23
4.1.2 breaks usage in Cloudflare Worker


leviwolfe:
Specifying `"faunadb": "4.1.1"` (along with the [known issue workaround](https://github.com/fauna/faunadb-js#using-with-cloudflare-workers)) in my `package.json` dependencies is working while `"faunadb": "^4.1.2"` unfortunately has the following error when attempting to create a faunadb client:

```"navigator is not defined"
ReferenceError: navigator is not defined
    at Object.getBrowserDetails (worker.js:1:2773)
    at worker.js:6:5260
    at new s (worker.js:6:5461)
    at new d (worker.js:6:17008)
    at new a (worker.js:6:33478)
    at s (worker.js:6:33984)
    at Proxy.e.r (worker.js:6:14246)
    at worker.js:6:34805
```

Sorry I'm not familiar enough with runtime de-obfuscating stack traces in Cloudflare Workers yet. I suspect this is new as of #436.

alizahid:
Also breaks when using with Next.js locally. However, seems to be working when deployed in production to Vercel. Although I have a feeling that it's pulling the older version from the cache when building in production.

My error, however, is different. I'm able to create a client and do queries, but when doing a query that creates a document, I get this error;

```
Error [ERR_HTTP2_GOAWAY_SESSION]: New streams cannot be created after receiving a GOAWAY
    at new NodeError (node:internal/errors:329:5)
    at ClientHttp2Stream.requestOnConnect (node:internal/http2/core:693:17)
    at node:internal/http2/core:1755:68
    at Array.forEach (<anonymous>)
    at ClientHttp2Session.<anonymous> (node:internal/http2/core:1755:11)
    at Object.onceWrapper (node:events:476:26)
    at ClientHttp2Session.emit (node:events:381:22)
    at emit (node:internal/http2/core:324:3)
    at processTicksAndRejections (node:internal/process/task_queues:83:21) {
  code: 'ERR_HTTP2_GOAWAY_SESSION'
}
```

RaymondMoay:
I'm having a similar error as @alizahid on NextJS. I can't do a simple read.

An API endpoint like this:

```
export default async (req, res) => {
    try {
      const doc = await db.query(
        q.Get(q.Ref(q.Collection("tweets"), req.query.id))
      );
      res.status(200).json(doc);
    } catch (e) {
      res.status(500).json({ error: e.message });
    }
  };
```

leads to an error like this:

`
{
[...]

PR #448, opened 2021-03-24
fix: cloudflare doesn't have navigator object

fireridlle:
### Notes
[DRV-566 Cloudflare doesn't have navigator object](https://faunadb.atlassian.net/browse/DRV-566)
Issue: https://github.com/fauna/faunadb-js/issues/447

Running 4.1.2 produce an error at CloudFlare
```
ReferenceError: navigator is not defined
    at Object.getBrowserDetails (worker.js:1:2773)
    at worker.js:6:5260
    at new s (worker.js:6:5461)
    at new d (worker.js:6:17008)
    at new a (worker.js:6:33478)
    at s (worker.js:6:33984)
    at Proxy.e.r (worker.js:6:14246)
    at worker.js:6:34805
```

### DIFF
--- a/src/_http/index.js
+++ b/src/_http/index.js
@@ -123,14 +123,24 @@ function getDefaultHeaders() {
     driver: ['javascript', packageJson.version].join('-'),
   }
 
+  let isServiceWorker
+
+  try {
+    isServiceWorker = global instanceof ServiceWorkerGlobalScope
+  } catch (error) {
+    isServiceWorker = false
+  }
+
   if (util.isNodeEnv()) {
     driverEnv.runtime = ['nodejs', process.version].join('-')
     driverEnv.env = util.getNodeRuntimeEnv()
     var os = require('os')
     driverEnv.os = [os.platform(), os.release()].join('-')
+  } else if (isServiceWorker) {
+    driverEnv.runtime = 'Service Worker'
   } else {
     driverEnv.runtime = util.getBrowserDetails()
-    driverEnv.env = 'unknown'
+    driverEnv.env = 'browser'
     driverEnv.os = getBrowserOsDetails()
   }
 
--- a/src/_http/index.js
+++ b/src/_http/index.js
@@ -131,18 +131,20 @@ function getDefaultHeaders() {
     isServiceWorker = false
   }
 
-  if (util.isNodeEnv()) {
-    driverEnv.runtime = ['nodejs', process.version].join('-')
-    driverEnv.env = util.getNodeRuntimeEnv()
-    var os = require('os')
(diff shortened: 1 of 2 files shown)


src/_http/index.js
```diff
@@ -123,17 +123,29 @@ function getDefaultHeaders() {
...
-  if (util.isNodeEnv()) {
-    driverEnv.runtime = ['nodejs', process.version].join('-')
-    driverEnv.env = util.getNodeRuntimeEnv()
-    var os = require('os')
-    driverEnv.os = [os.platform(), os.release()].join('-')
-  } else {
-    driverEnv.runtime = util.getBrowserDetails()
-    driverEnv.env = 'unknown'
-    driverEnv.os = getBrowserOsDetails()
+  let isServiceWorker
```
erickpintor, line 126: Is let ECMA 5? The rest of the codebase uses `var`.

src/_http/index.js
```diff
@@ -123,17 +123,29 @@ function getDefaultHeaders() {
...
+      var os = require('os')
+      driverEnv.os = [os.platform(), os.release()].join('-')
+    } else if (isServiceWorker) {
+      driverEnv.runtime = 'Service Worker'
+    } else {
+      driverEnv.runtime = util.getBrowserDetails()
+      driverEnv.env = 'browser'
+      driverEnv.os = getBrowserOsDetails()
+    }
+  } catch (_) {}
```
erickpintor, line 147: Same here. Is this syntax ECMA 5 compatible?

fireridlle: yes. to be honest, the driver isn't es5 compatible. just an example, promises is es6

erickpintor: It used to be a policy that JS driver was supposed be ECMA5 compatible. It had a pollyfill for promises, I believe. If this is not the case anymore, we should review the policy has changed or if we introduced a regression.

fireridlle: driver doesn't have pollfyfill for promises. even more, doc said that it's es6 promises
https://github.com/fauna/faunadb-js#instantiating-a-client-and-issuing-queries

also, we are working on modularization which would have bundle tools from es6+ to anything, including es5.


https://caniuse.com/?search=es6
according to this table, only pretty old browsers doesn't support es6

erickpintor: I don't disagree and I'm fine moving to es6. Let's just make sure this is done via a conscious and well documented decision rather than through a non-backwards compatible change.

parkhomenko (approved):

erickpintor (approved):
LGTM

PR #453, opened 2021-03-26
feature: split each function in query.js to separate file

fireridlle:
### Notes
[Split “query.js” single file into separate files with ESM exports](https://faunadb.atlassian.net/browse/DRV-555)

Split each function inside query.js to separate file with ESM exports
Update Readme.md with migrations guide
Support for NodeJS esm


src/query/Lambda.js
```diff
@@ -0,0 +1,50 @@
+import { InvalidValue } from '../errors'
+import Expr from '../Expr'
+import { checkInstanceHasProperty } from '../_util'
+import arity from './arity'
+import lambdaExpr from './lambdaExpr'
+import lambdaFunc from './lambdaFunc'
```
marrony, line 6: is there any benefit of extracting `lambdaExpr` and `lambdaFunc` in separate files? I think those functions are only used here

src/query/wrap.js
```diff
@@ -0,0 +1,66 @@
...
+/**
+ * Wraps an object as an Expression. This will automatically wrap any bare objects with
+ * the appropriate {@link object} escaping.
+ * @param {Object} obj
+ *  The object to be wrapped as an Expression.
+ * @returns {Expr}
+ *   The expression wrapping the provided object.
+ * @private
+ */
+export function wrap(obj) {
```
marrony, line 16: why not put all those utility functions (like `wrap`, `varargs`, `params` and &c) into a single file?

tigger9flow (approved):

PR #456, opened 2021-03-31
fix: cross platform global

fireridlle:
Original discussion https://github.com/fauna/faunadb-js/pull/428

Cross-platform global object (globalThis, window, self, global)

### DIFF
--- a/src/_util.js
+++ b/src/_util.js
@@ -451,6 +451,12 @@ function mergeObjects(obj1, obj2) {
   return obj3
 }
 
+/**
+* globalThis - cross platform this, but IE doesn't support it
+* window - browser global context
+* global - nodejs global context
+* self - service worker global context
+*/
 var crossGlobal =
   typeof window !== 'undefined'
     ? window
--- a/src/_util.js
+++ b/src/_util.js
@@ -51,6 +51,25 @@ function isNodeEnv() {
 }
 
 /**
+ * Resolves environment variable if available.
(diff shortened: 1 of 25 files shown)


tigger9flow (dismissed):

src/_util.js
```diff
@@ -451,7 +451,17 @@ function mergeObjects(obj1, obj2) {
   return obj3
 }
 
+var crossGlobal =
```
erickpintor, line 454: Could you maybe add some docs here explaining the motivation for this? Abdicating from using global `window` variable might be counter intuitive without further context.

fireridlle: almost all platforms supports globalThis, which is kinda built-in polyfill for global context to cover cross platforms. but we can't use it as IE11 doesn't support it https://caniuse.com/?search=globalThis

window - browser global context
global - nodejs global context
self - service worker global context

so crossGlobal will contain one of them depends on availability

erickpintor: That's good information, would you mind adding that as a comment to the code?

parkhomenko (dismissed):

erickpintor (dismissed):
🐑 it.

tigger9flow (dismissed):

parkhomenko (approved):

parkhomenko (approved):

tigger9flow (approved):

PR #457, opened 2021-03-31
fix: use npm for test running

fireridlle:
Repo contains package-lock.json, however, yarn is used to run tests on circleci. 
It is advised not to mix package managers in order to avoid resolution inconsistencies caused by unsynchronized lock files.


tigger9flow (approved):

faunaee (approved):
LGTM! 👍

PR #460, opened 2021-04-05
DRV-592 - Experiment with changing the way the http2 session timeout works [DRAFT]

retwere:
### Notes
[Jira Ticket]()

### Screenshots

### DIFF
--- a/src/_http/http2Adapter.js
+++ b/src/_http/http2Adapter.js
@@ -3,8 +3,8 @@ var http2 = require('http2')
 var errors = require('./errors')
 var util = require('../_util')
 
-// Destroy session after 1 minute of inactivity.
-var DESTROY_HTTP2_SESSION_TIME = 1000 * 60
+// Destroy session after 0.5 seconds of inactivity.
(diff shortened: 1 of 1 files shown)


faunaee:
I just tried these changes, and they seem to work well! I did not get any of the `GOAWAY` errors encountered with the previous timeout strategy, and I could set the timeout down to 0 to run all of the examples in the documentation as quickly as possible.

Two suggestions:

1) Allow specification of `DESTROY_HTTP2_SESSION_TIME` via an environment variable:
   ```
   var DESTROY_HTTP2_SESSION_TIME = process.env('FAUNADB_HTTP_SESSION_TIMEOUT') || 500
   ```

   That would allow a developer to set the timeout to 0 to provide similar performance/behaviour as the 4.0.3 driver without requiring a code change in their clients that use the JS driver.

2) Reinstate the `http2SessionIdleTime` option for client instantiation from the 4.1.2 so that developers can set the timeout per client connection.

retwere:
Closing this since it was done in https://github.com/fauna/faunadb-js/pull/461.

PR #461, opened 2021-04-06
DRV-592 - implement inactivity timeout for HTTP2 session

tigger9flow:
### Notes
[Jira Ticket](https://faunadb.atlassian.net/browse/DRV-592)

### How to test
Can be tested by writing a simple script with a fauna query and assert that after its execution fauna `Client` doesn't hold the event loop (the script is terminated after 500ms).
The following scripts have been used to load test inactivity mechanics by running parallel queries.
Perform `5000` parallel queries over `10` randomly selected `Client` instances:
```typescript
// inactivityTest.ts
require('dotenv').config()
import { Client, query as Q } from 'faunadb'

const SECRET = process.env.FAUNA_SECRET_ADMIN_KEY!

const getRandomInt = (min: number, max: number): number => {
  min = Math.ceil(min)
  max = Math.floor(max)

  return Math.floor(Math.random() * (max - min + 1)) + min
}

const getRandomElement = <T>(list: T[]): T =>
  list[getRandomInt(0, list.length - 1)]

const makeClient = () =>
  new Client({
    secret: SECRET,
    domain: 'db.fauna-preview.com',
    scheme: 'https',
  })

const clientPool = Array.from({ length: 10 }, makeClient)

for (const _ of '_'.repeat(5000)) {
  const client = getRandomElement(clientPool)
  const queries = [
    // Paginate query
    Q.Map(Q.Paginate(Q.Documents(Q.Collection('items'))), ref =>
      Q.Get(ref),
    ),
    // Simple query
    Q.Sum([1, 1]),
  ]

  client.query(getRandomElement(queries)).catch(err => {
    console.error(err)
    process.exit(1)
  })
}
```

To simulate not only parallel queries but also a random delay between them:
```typescript
// inactivityTestWithDelay.ts
require('dotenv').config()
import { Client, query as Q } from 'faunadb'

const IN_PARALLEL = 500
const MAX_ATTEMPTS = 10
const SECRET = process.env.FAUNA_SECRET_ADMIN_KEY!

const getRandomInt = (min: number, max: number): number => {
  min = Math.ceil(min)
  max = Math.floor(max)

  return Math.floor(Math.random() * (max - min + 1)) + min
}

const getRandomElement = <T>(list: T[]): T =>
  list[getRandomInt(0, list.length - 1)]

const makeClient = () =>
  new Client({
    secret: SECRET,
    domain: 'db.fauna-preview.com',
    scheme: 'https',
  })

const clientPool = Array.from({ length: 10 }, makeClient)

const tick = () => {
  const id = Math.random().toString().slice(-5)
  const requests: Promise<unknown>[] = []

  console.log(`Running ${IN_PARALLEL} parallel queries for ${id}`)

  for (const _ of '_'.repeat(IN_PARALLEL)) {
    const client = getRandomElement(clientPool)
    const queries = [
      Q.Map(Q.Paginate(Q.Documents(Q.Collection('items'))), ref =>
        Q.Get(ref),
      ),
      Q.Sum([1, 1]),
    ]

    requests.push(
      client.query(getRandomElement(queries)).catch(err => {
        console.error(err)
        process.exit(1)
      }),
    )
  }

  Promise.all(requests).then(() => console.log(`Done for ${id}`))
}

const execute = (idx = MAX_ATTEMPTS) => {
  const delay = idx === MAX_ATTEMPTS ? 0 : getRandomInt(0, 10 * 1000)

  setTimeout(() => {
    tick()

    if (idx !== 0) {
      execute(idx - 1)
    }
  }, delay)
}

execute()
```

### DIFF
--- a/src/Client.js
+++ b/src/Client.js
@@ -155,9 +155,21 @@ var values = require('./values')
  * @param {?fetch} options.fetch
  *   a fetch compatible [API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) for making a request
  * @param {?number} options.queryTimeout
- *   Sets the maximum amount of time (in milliseconds) for query execution on the server,
+ *   Sets the maximum amount of time (in milliseconds) for query execution on the server
+ * @param {?number} options.http2SessionIdleTime
+ *   Sets the maximum amount of time (in milliseconds) an HTTP2 session may live
+ *   when there's no activity. Only applicable for NodeJS environment (when http2 module is used), 500 by default,
+ *   can be configured via FAUNADB_HTTP2_SESSION_IDLE_TIME environment variable.
  */
 function Client(options) {
+  var http2SessionIdleTimeEnv = parseInt(
+    util.getEnvVariable('FAUNADB_HTTP2_SESSION_IDLE_TIME'),
+    10
+  )
+  var http2SessionIdleTimeDefault = !isNaN(http2SessionIdleTimeEnv)
+    ? http2SessionIdleTimeEnv
+    : 500
+
   options = util.applyDefaults(options, {
     domain: 'db.fauna.com',
     scheme: 'https',
@@ -169,6 +181,7 @@ function Client(options) {
     headers: {},
     fetch: undefined,
     queryTimeout: null,
+    http2SessionIdleTime: http2SessionIdleTimeDefault,
   })
 
   this._observer = options.observer
--- a/src/_http/http2Adapter.js
+++ b/src/_http/http2Adapter.js
@@ -3,17 +3,18 @@ var http2 = require('http2')
 var errors = require('./errors')
 var util = require('../_util')
 
-// Destroy session after 1 minute of inactivity.
-var DESTROY_HTTP2_SESSION_TIME = 1000 * 60
 var STREAM_PREFIX = 'stream::'
 
 /**
  * Http client adapter built around NodeJS http2 module.
  *
  * @constructor
+ * @param {object} options Http2Adapter options.
+ * @param {number} options.http2SessionIdleTime The time (in milliseconds) that 
+ * an HTTP2 session may live when there's no activity.
  * @private
  */
-function Http2Adapter() {
+function Http2Adapter(options) {
   /**
    * Identifies a type of adapter.
    *
@@ -27,6 +28,13 @@ function Http2Adapter() {
    * @private
    */
   this._sessionMap = {}
+  /**
+   * The time (in ms) that an HTTP2 session may live when there's no activity.
+   *
+   * @type {number}
+   * @private
+   */
+  this._http2SessionIdleTime = options.http2SessionIdleTime
 }
 
 /**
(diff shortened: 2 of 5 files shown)


src/Client.js
```diff
@@ -169,6 +172,7 @@ function Client(options) {
     headers: {},
     fetch: undefined,
     queryTimeout: null,
+    http2SessionIdleTime: 500,
```
retwere, line 175: Have you read Ewan's comment on my PR? https://github.com/fauna/faunadb-js/pull/460#issuecomment-813687728

What do you think? Can we also allow setting the default idle time by environment variable?

src/_http/http2Adapter.js
```diff
@@ -36,29 +44,68 @@ function Http2Adapter() {
  * @param {?boolean} isStreaming Whether it's a streaming request. A separate session
  * is created for streaming requests to avoid shared resources with regular
  * ones for the purpose of reliability.
- * @returns {ClientHttp2Session} Http2 session.
+ * @returns {object} Http2 session API.
```
retwere, line 47: nit: I'm not a huge fan of "session API" as a name for this object. "API" makes me think of something exterior-facing.

Maybe we can call it a "sessionAdapter" instead (to match the Http2Adapter)?

retwere (approved):
LGTM. See minor comments below.

Can you also check in the load tests (along with instructions on how to run them)? It will be much easier to find them and run them again in the future if they're checked in.

faunaee (commented):
I added a suggestion for defining the default timeout via environment variable.

src/_http/http2Adapter.js
tigger9flow: @retwere Good catch, it's confusing. I've named it API as it's supposed to be like an interface to work with a session. How do you feel about "sessionInterface"?

src/Client.js
```diff
@@ -155,7 +155,10 @@ var values = require('./values')
...
  *   a fetch compatible [API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) for making a request
  * @param {?number} options.queryTimeout
- *   Sets the maximum amount of time (in milliseconds) for query execution on the server,
+ *   Sets the maximum amount of time (in milliseconds) for query execution on the server
+ * @param {?number} options.http2SessionIdleTime
+ *   Sets the maximum amount of time (in milliseconds) an HTTP2 session may live
+ *   when there's no activity. Only applicable for NodeJS environment (when http2 module is used), 500 by default.
  */
 function Client(options) {
   options = util.applyDefaults(options, {
```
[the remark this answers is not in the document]
tigger9flow: @faunaee it looks good, I like the idea, especially if it could fix your docs tests and avoid the delay. One concern is that we need to verify if it's NodeJS env to access the "process" property but no worries, I'll fix and test it tomorrow. Thank you, Ewan!

src/_http/http2Adapter.js
retwere: Sounds good to me 👍

src/Client.js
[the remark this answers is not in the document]
tigger9flow: @faunaee updated to pull from env variable by default. I use FAUNADB_HTTP2_SESSION_IDLE_TIME name for the env variable to match `Client`'s option name

src/Client.js
```diff
@@ -155,9 +155,21 @@ var values = require('./values')
  * @param {?fetch} options.fetch
  *   a fetch compatible [API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) for making a request
  * @param {?number} options.queryTimeout
- *   Sets the maximum amount of time (in milliseconds) for query execution on the server,
+ *   Sets the maximum amount of time (in milliseconds) for query execution on the server
+ * @param {?number} options.http2SessionIdleTime
+ *   Sets the maximum amount of time (in milliseconds) an HTTP2 session may live
+ *   when there's no activity. Only applicable for NodeJS environment (when http2 module is used), 500 by default,
+ *   cab be configured via FAUNADB_HTTP2_SESSION_IDLE_TIME environment variable.
```
retwere, line 162: typo: `cab` should be `can`
```suggestion
 *   can be configured via FAUNADB_HTTP2_SESSION_IDLE_TIME environment variable.
```

src/_http/http2Adapter.js
```diff
@@ -3,17 +3,18 @@ var http2 = require('http2')
...
-var DESTROY_HTTP2_SESSION_TIME = 1000 * 60
 var STREAM_PREFIX = 'stream::'
 
 /**
  * Http client adapter built around NodeJS http2 module.
  *
  * @constructor
+ * @param {object} options Http2Adapter options.
+ * @param {number} options.http2SessionIdleTime The time an HTTP2 session
+ * may live when there's no activity.
```
retwere, line 14: add `(in milliseconds)`
```suggestion
 * @param {number} options.http2SessionIdleTime The time (in milliseconds) that 
 * an HTTP2 session may live when there's no activity.
```

src/_http/http2Adapter.js
```diff
@@ -27,6 +28,13 @@ function Http2Adapter() {
    * @private
    */
   this._sessionMap = {}
+  /**
+   * The time an HTTP2 session may live when there's no activity.
```
retwere, line 32: add `(in  ms)`
```suggestion
   * The time (in ms) that an HTTP2 session may live when there's no activity.
```

retwere (dismissed):
LGTM

retwere:
Also, please check the load tests in somewhere, if not in this PR, then in a separate one. I would hate to lose track of the work you've done to validate the change.

tigger9flow:
> Also, please check the load tests in somewhere, if not in this PR, then in a separate one. I would hate to lose track of the work you've done to validate the change.

Thanks! Tests are supposed to be added in scope of [this ticket](https://faunadb.atlassian.net/browse/DRV-562). I think it may be separate npm script, `npm run test:load`

retwere (approved):
👍

parkhomenko (approved):

PR #471, opened 2021-04-13
DRV-575 - implement Client#close method

tigger9flow:
* implement the `Client#close` method to manually clean up an HTTP2 session
* allow `Infinity` for `http2SessionIdleTime` parameter

### Notes
[Jira Ticket](https://faunadb.atlassian.net/browse/DRV-575)

### How to test
The following script has been used to test the proper termination:
```typescript
// sample.ts
import { Client, query as Q } from 'faunadb'

const SECRET = process.env.FAUNA_SECRET_ADMIN_KEY!

const client = new Client({
  secret: SECRET,
  domain: 'db.fauna-preview.com',
  scheme: 'https',
  http2SessionIdleTime: Infinity,
})

client.query(Q.Now()).then(console.log)

setTimeout(() => {
  console.log('Force closing...')
  client.close()
}, 5 * 1000)
```

* run `ts-node sample.ts`
* assert script is terminated after around 5secs

### DIFF
--- a/README.md
+++ b/README.md
@@ -242,6 +242,45 @@ const client = new faunadb.Client({
 })
 ```
 
+#### HTTP/2 Session Idle Time (Node.js only)
+
+When running on the Node.js platform, the Fauna client uses [HTTP/2 multiplexing](https://stackoverflow.com/questions/36517829/what-does-multiplexing-mean-in-http-2)
+to reuse the same session for many simultaneous requests. After all open requests
+have been resolved, the client will keep the session open for a period of time
+(500ms by default) to be reused for any new requests.
+
+The `http2SessionIdleTime` parameter may be used to control how long the HTTP/2
+session remains open while the connection is idle. To save on the overhead of
+closing and re-opening the session, set `http2SessionIdleTime` to a longer time
+--- or even `Infinity`, to keep the session alive indefinitely.
+
+While an HTTP/2 session is alive, the client will hold the Node.js event loop
+open; this prevents the process from terminating. Call `Client#close` to manually
+close the session and allow the process to terminate. This is particularly
+important if `http2SessionIdleTime` is long or `Infinity`:
+
+```javascript
+// sample.js (run it with "node sample.js" command)
+const { Client, query: Q } = require('faunadb')
+
+async function main() {
+  const client = new Client({
+    secret: 'YOUR_FAUNADB_SECRET',
+    http2SessionIdleTime: Infinity,
+    //                    ^^^ Infinity or non-negative integer
+  })
+  const output = await client.query(Q.Add(1, 1))
+
+  console.log(output)
+
+  client.close()
+  //     ^^^ If it's not called then the process won't terminate
+}
+
+main().catch(console.error)
+```
+
+
 ## Known issues
 
 ### Using with Cloudflare Workers
--- a/src/Client.js
+++ b/src/Client.js
@@ -158,17 +158,14 @@ var values = require('./values')
  *   Sets the maximum amount of time (in milliseconds) for query execution on the server
  * @param {?number} options.http2SessionIdleTime
  *   Sets the maximum amount of time (in milliseconds) an HTTP2 session may live
- *   when there's no activity. Only applicable for NodeJS environment (when http2 module is used), 500 by default,
- *   can be configured via FAUNADB_HTTP2_SESSION_IDLE_TIME environment variable.
+ *   when there's no activity. Must either be a non-negative integer, or Infinity to allow the
+ *   HTTP2 session to live indefinitely (use `Client#close` to manually terminate the client).
+ *   Only applicable for NodeJS environment (when http2 module is used). Default is 500ms;
+ *   can also be configured via the FAUNADB_HTTP2_SESSION_IDLE_TIME environment variable
+ *   which has the highest priority and overrides the option passed into the Client constructor.
  */
 function Client(options) {
-  var http2SessionIdleTimeEnv = parseInt(
-    util.getEnvVariable('FAUNADB_HTTP2_SESSION_IDLE_TIME'),
-    10
-  )
-  var http2SessionIdleTimeDefault = !isNaN(http2SessionIdleTimeEnv)
-    ? http2SessionIdleTimeEnv
-    : 500
+  var http2SessionIdleTime = getHttp2SessionIdleTime()
 
   options = util.applyDefaults(options, {
     domain: 'db.fauna.com',
@@ -181,8 +178,12 @@ function Client(options) {
     headers: {},
     fetch: undefined,
     queryTimeout: null,
-    http2SessionIdleTime: http2SessionIdleTimeDefault,
+    http2SessionIdleTime: http2SessionIdleTime.value,
   })
+
+  if (http2SessionIdleTime.shouldOverride) {
+    options.http2SessionIdleTime = http2SessionIdleTime.value
+  }
 
   this._observer = options.observer
   this._http = new http.HttpClient(options)
@@ -259,6 +260,24 @@ Client.prototype.syncLastTxnTime = function(time) {
   this._http.syncLastTxnTime(time)
 }
 
+/**
+ * Closes the client session and cleans up any held resources.
+ * By default, it will wait for any ongoing requests to complete on their own;
+ * streaming requests are terminated forcibly. Any subsequent requests will
+ * error after the .close method is called.
+ * Should be used at application termination in order to release any open resources
+ * and allow the process to terminate e.g. when the custom http2SessionIdleTime parameter is used.
+ *
+ * @param {?object} opts Close options.
+ * @param {?boolean} opts.force Specifying this property will force any ongoing
+ * requests to terminate instead of gracefully waiting until they complete.
+ * This may result in an ERR_HTTP2_STREAM_CANCEL error for NodeJS.
+ * @returns {Promise<void>}
+ */
+Client.prototype.close = function(opts) {
+  return this._http.close(opts)
+}
+
 Client.prototype._execute = function(method, path, data, query, options) {
   query = util.defaults(query, null)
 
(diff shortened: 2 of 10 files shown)


tigger9flow:
@faunaee I'd appreciate your review of README and JSDoc

parkhomenko (dismissed):

erickpintor (approved):
LGTM

retwere (changes requested):
Even though we're adding the `close` method to solve a specific problem (application being kept alive too long by HTTP2 sessions), we should also make sure that 'closing' a client is a coherent concept, that a user can make intuitive sense of. Most users will consider HTTP2 session timeouts to be an implementation detail; but they will be used to the idea of a client keeping resources open that need to be cleaned up. I think there's a few ways we could make our `close` method better conform to user's expectations. See below for detailed comments.

Also, you asked @faunaee for comments on the docs, but I couldn't help but leave a few. Hope you don't mind. :)

src/Client.js
```diff
@@ -327,4 +338,18 @@ Client.prototype._handleRequestResult = function(response, result, options) {
...
 
+function getHttp2SessionIdleTime() {
+  var fromEnv = util.getEnvVariable('FAUNADB_HTTP2_SESSION_IDLE_TIME')
+  var parsed =
+    // Allow either "Infinity" or parsable integer string.
+    fromEnv === 'Infinity' ? Infinity : parseInt(fromEnv, 10)
+  var useEnvVar = !isNaN(parsed)
+
+  return {
+    // Env variable has the highest priority.
```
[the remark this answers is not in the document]
tigger9flow: @retwere In my experience, it works this way to leave the code as-is if something needs to be configured externally. But let me know if there're any reasons we should avoid that. Will add this to the description to make it clear.

src/_http/http2Adapter.js
```diff
@@ -279,4 +290,17 @@ Http2Adapter.prototype.execute = function(options) {
...
 }
 
+/**
+ * Destroys all the sessions and resets the session map.
+ *
+ * @returns {void}
+ */
+Http2Adapter.prototype.close = function() {
+  Object.values(this._sessionMap).forEach(function(sessionInterface) {
+    sessionInterface.destroy()
```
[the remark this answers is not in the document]
tigger9flow: Thanks for your help, greatly appreciate your input! You're totally right that it can be dangerous to destroy sessions expecting that the user will take care that there's no ongoing traffic and avoid any consequences caused by the possible interruption.
I've tested with the `.close` method and it works fine for regular requests (at least when they don't take much time to complete). But it hangs for streaming ones as expected, the script below won't be terminated:
```typescript
import { Client, query as Q } from 'faunadb'

const SECRET = process.env.FAUNA_SECRET_ADMIN_KEY!
const DOC_REF = Q.Ref(Q.Collection('items'), '295854750489903619')

const client = new Client({
  secret: SECRET,
  domain: 'db.fauna-preview.com',
  scheme: 'https',
})

async function main() {
  await new Promise<void>((resolve, reject) => {
    client.stream
      .document(DOC_REF)
      .on('snapshot', () => resolve())
      .on('error', reject)
      .start()
  })

  console.log("Listen to doc's changes")

  setTimeout(() => {
    console.log('Closing...')
    client.close()
  }, 2000)
}

main()
```

How do you think about `.close`ing regular requests and `.destroy`ing streaming ones? It may be some sort of compromise solution assuming that interruption of a streaming request isn't that crucial compared to a regular one which can perform a very important query.

We could also allow a parameter for the `Client#close` to leverage `.destroy` API instead of `.close`, e.g.:
```javascript
client.close({
  force: true, // Points to use session.destroy instead
})
```

README.md
```diff
@@ -242,6 +242,42 @@ const client = new faunadb.Client({
 })
 ```
 
+#### Termination
```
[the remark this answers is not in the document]
tigger9flow: @retwere Makes sense, I've changed the section title to "HTTP2 Session Idle Time (NodeJS only)".

> it might be good to split this discussion into two sections, one for termination...

I couldn't figure out what we can put into the "Termination" section as it only makes sense in the context of `http2SessionIdleTime` for now and is directly related to the "HTTP2 Session Idle Time" section.

src/_http/http2Adapter.js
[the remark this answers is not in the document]
tigger9flow: @retwere  I've pushed new changes to conform to what you've listed above, please check it out and let me know whether it makes sense.

Unfortunately, there is still a chance to hit the `ERR_HTTP2_GOAWAY_SESSION` error by calling the `.close` method and I have no idea what to do about it. The error may occur if the user calls `Client#close` too quickly (for me it's reproducible for 200ms delay) like on that simple code snippet we started our discussion from.
Here's a code snippet to reproduce on the latest PR's version:
```typescript
import { Client, query as Q } from 'faunadb'

const SECRET = process.env.FAUNA_SECRET_ADMIN_KEY!

const client = new Client({
  secret: SECRET,
  domain: 'db.fauna-preview.com',
  scheme: 'https',
})

function main() {
  client
    .query(Q.Now())
    .then(val => console.log('Response:', val))
    .catch(err => console.error('Failed:', err))

  setTimeout(() => {
    client.close()
  }, 200)
}

main()
```

test/client.test.js
```diff
@@ -143,6 +143,29 @@ describe('Client', () => {
     return expect(resultWithoutOptions).toEqual(resultWithOptions)
   })
 
+  test('cleans up http2 sessions when .close is called', async () => {
```
retwere, line 146: Any way we can test the behavior for both HTTP1 and HTTP2? We have changed the behavior for HTTP1 so we should test the new semantics (no requests after close).

test/client.test.js
```diff
@@ -143,6 +143,29 @@ describe('Client', () => {
...
+    const client = util.getClient({
+      http2SessionIdleTime: Infinity,
+    })
+
+    await client.ping()
+
+    expect(client._http._adapter._closed).toBe(false)
+
+    await client.close()
```
retwere, line 155: Can we add a test case with force=true?

src/Client.js
```diff
@@ -259,6 +260,24 @@ Client.prototype.syncLastTxnTime = function(time) {
...
+ * Closes the client session and cleans up any held resources.
+ * By default, it will wait for any ongoing requests to complete on their own;
+ * streaming requests are terminated forcibly. Any subsequent requests will
+ * error after the .close method is called.
+ * Should be used at application termination in order to release any open resources
+ * and allow the process to terminate e.g. when the custom http2SessionIdleTime parameter is used.
+ *
+ * @param {?object} opts Close options.
+ * @param {?boolean} opts.force Specifying this property will force any ongoing
+ * requests to terminate instead of gracefully waiting until they complete.
```
retwere, line 273: Currently not true for the implementation of HTTP1. Either the docs or implementation should change. My preference would be for implementation to match this description -- it's easier for users to understand if the behavior is consistent (HTTP1 vs. HTTP2 is an implementation detail that most users won't care much about).

retwere (changes requested):
A few more comments, and again I couldn't help but make a bunch of suggestions on the documentation.

I'll take a look at the GOAWAY error case you described. If we can't figure out how to prevent it then we should catch the error and wrap it in an Error class of our own. I think we've concluded that the only way the GOAWAY error can occur is if the session is closed when the request initiates, so maybe something like `ClientClosed`. At least we could give a better error message than the default Node.js one.

parkhomenko (approved):

tigger9flow:
@retwere Hi, Jake! I've pushed new changes to this PR. Implemented the force option for HTTP/1 and resolving .close's Promise only when all of the requests complete. It a bit complicated implementation but now HTTP/1 should match HTTP/2, except the fact HTTP/2 throws if `.close` is called immediately or with a [small delay](https://github.com/fauna/faunadb-js/pull/471#discussion_r614805159). Now [it remaps](https://github.com/fauna/faunadb-js/pull/471/files#diff-5deaa864c6c703bfed3ed33ae4c72fd57b8367abeab8fae5ca08ebf9a2b67b82R362) `ERR_HTTP2_GOAWAY_SESSION` error into `ClientClosed` one, feel free to suggest more reasonable error message. I'll be on vacation until May 11th, would love to hear your feedback when I return. :)

parkhomenko (dismissed):

fireridlle (dismissed):

src/_http/http2Adapter.js
```diff
@@ -279,4 +325,46 @@ Http2Adapter.prototype.execute = function(options) {
...
+
+  this._closed = true
+
+  var noop = function() {}
+
+  return Promise.all(
+    Object.values(this._sessionMap).map(function(sessionInterface) {
+      return sessionInterface.close(opts.force)
+    })
+  ).then(noop)
```
retwere, line 346: I'm curious why the noop is needed?

test/client.test.js
```diff
@@ -143,6 +147,97 @@ describe('Client', () => {
...
+    const client = util.getClient()
+
+    await expect(
+      Promise.all([
+        client.ping(),
+        client.close({
+          force: true,
+        }),
+      ])
+    ).rejects.toThrow('The request is aborted due to the Client#close call')
```
retwere, line 183: Is there any way to check the class (`ClientClosed`) rather than the error message? It would be nice if the test did not depend on the particular phrasing of the message.

retwere (approved):
LGTM. Made a couple of minor comments but changes are not necessary to merge. 

🚀

src/_http/http2Adapter.js
tigger9flow: @retwere It forces returning `Promise<void>` instead of `Promise<undefined[]>` to be consistent and conform to JS doc definition, otherwise:
```typescript
Promise.all([Promise.resolve()]) // Returns Promise<undefined[]>
```
If we used async/await, we could just "await" it and return nothing:
```typescript
  ...
  await Promise.all(...)
}
```

parkhomenko (approved):

fireridlle (approved):

PR #474, opened 2021-04-19
feature: load test

fireridlle:
### Notes
[Load test](https://faunadb.atlassian.net/browse/DRV-562)
Before release, we need to check if the driver can perform under production load usage.
Create multiple clients and run multiple concurrent queries

```
npm run load-test -- --help

Options:
      --help     Show help                                             [boolean]
      --version  Show version number                                   [boolean]
  -t, --threads  Maximum number of consequence threads    [number] [default: 10]
  -q, --queries  Maximum number of queries per each thread [number] [default: 500]
```

### DIFF
--- a/tools/loadTest.js
+++ b/tools/loadTest.js
@@ -1,0 +1,76 @@
+const yargs = require('yargs')
+const faunadb = require('../')
+const testConfig = require('../test/config')
+const { Client, query: Q } = faunadb
+
+const { argv } = yargs
+  .option('threads', {
+    alias: 't',
+    description: 'Maximum number of consequence threads',
+    type: 'number',
+    default: 10,
+  })
+  .option('queries', {
+    alias: 'q',
+    description: 'Maximum number of queries per each thread',
+    type: 'number',
+    default: 500,
+  })
+
+const getRandomInt = (min, max) => {
+  min = Math.ceil(min)
+  max = Math.floor(max)
+
+  return Math.floor(Math.random() * (max - min + 1)) + min
+}
+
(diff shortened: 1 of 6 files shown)


tigger9flow (approved):
LGTM
Please, don't forget to mention this script in the release instruction (confluence)

tools/loadTest.js
```diff
@@ -0,0 +1,76 @@
...
+}
+
+const getRandomElement = list => list[getRandomInt(0, list.length - 1)]
+
+const makeClient = () =>
+  new Client({
+    secret: testConfig.auth,
+    domain: testConfig.domain,
+    scheme: testConfig.scheme,
+  })
```
evbruno, line 34:
```suggestion
    port: testConfig.port
  })
```

evbruno: I had to add the port to run the `load-test` locally (inside a docker container); 
Is it supposed to catch the `FAUNA_PORT` (env var) from other place?

evbruno: In the end I got a few `(node:171) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 connect listeners added. Use emitter.setMaxListeners() to increase limit` but I think the PR is good to go

fireridlle: the approach of getting config is the same as used to be. driver tries to read file `testConfig` at the root dir. If failed, he tries to get config from env

evbruno (approved):

parkhomenko (approved):

PR #365, opened 2020-11-18
fix: updates to stream types

ecwyne:
Types for `client.stream()` callback handlers had incorrect number of arguments. This PR also adds types for `client.stream.document()` and correctly allows `snapshot` events in `client.stream.document()` and not in `client.stream()`

PR #478, opened 2021-04-27
WIP: 5.x.x

fireridlle:
### Notes
[Jira Ticket]()

### Screenshots


steida:
For The Love Of God, is it so hard to just use TypeScript? It's 2021. Please! People deserve types, especially with FQL 😭

fireridlle:
Migrating to typescript in our backlog. But as this process is not as fast as we would like to because we need not only provide good public typing but also typing for all internal functions and structures. Therefore, we decided to do it iteratively, and as a first step, we provide EMS. The main goal of ESM is to enable tree shaking.

Although current types of coverage are pure, you can use faunadb-js with typescript. Let us know if you have any types of issues or feel free to contribute. We are going to consolidate https://github.com/fauna/faunadb-js/pull/252 and https://github.com/fauna/faunadb-js/pull/365 and other types related changes soon

PR #482, opened 2021-04-29
fix: check if user in organization

fireridlle:
Step to check if a user in the organization return string type rather than boolean type


vadimLF (approved):
LGTM.

shestakovg (approved):

PR #507, opened 2021-06-22
fix(OSS-800): enchange error message

fireridlle:
### Notes
[Fix false positive cases for ClientClosed error](https://faunadb.atlassian.net/browse/OSS-800)

### DIFF
--- a/src/_http/http2Adapter.js
+++ b/src/_http/http2Adapter.js
@@ -200,7 +200,7 @@ Http2Adapter.prototype.execute = function(options) {
     // we need to call streamConsumer.onError instead of reject function.
     // Possible scenario is aborting request when stream is already being consumed.
     var rejectOrOnError = function(error) {
-      var remapped = remapHttp2Error(error)
+      var remapped = remapHttp2Error({ error, isClosed: self._closed })
 
       if (isPromiseSettled && isStreaming) {
         return options.streamConsumer.onError(remapped)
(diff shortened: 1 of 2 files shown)


parkhomenko (approved):

src/_http/http2Adapter.js
```diff
@@ -353,10 +353,11 @@ Http2Adapter.prototype.close = function(opts) {
  * @param {Error} error Error object.
  * @returns {Error} Remapped error.
  */
-function remapHttp2Error(error) {
+function remapHttp2Error({ error, isClosed }) {
   var shouldRemap =
-    error.code === 'ERR_HTTP2_GOAWAY_SESSION' ||
-    error.code === 'ERR_HTTP2_STREAM_CANCEL'
+    !isClosed &&
```
tigger9flow, line 358: We should remap to `ClientClosed` error only if `isClosed` is `true`:
```suggestion
    isClosed &&
```

test/client.test.js
```diff
@@ -173,14 +173,10 @@ describe('Client', () => {
...
 
-    await expect(
-      Promise.all([
-        client.ping(),
-        client.close({
-          force: true,
-        }),
-      ])
-    ).rejects.toThrow(errors.ClientClosed)
+    await client.close({
```
tigger9flow, line 176: If guess tests should be reverted as well. In this test case, we test parallel execution of `.close` and `.ping`

tigger9flow (approved):

parkhomenko (approved):

PR #535, opened 2021-09-10
Tests/oss 904 v4/v5 pipelines

fireridlle:


concourse/pipeline-v4.yml
```diff
@@ -0,0 +1,47 @@
+---
+resources:
+  - name: fauna-js-repository
+    type: git
+    icon: github
+    source:
+      uri: https://github.com/fauna/faunadb-js.git
+      branch: main
```
parkhomenko, line 8: I think this should be `v4` branch

concourse/tasks/integration.yml
```diff
@@ -10,20 +29,18 @@ services:
...
-    image: node:15.14.0-alpine3.10
-    container_name: mytests
-    depends_on:
-      - faunadb
-    volumes:
-      - "../../:/tmp/app"
-    working_dir: "/tmp/app"
-    command: [sh, -c, "npm install && npm test"]
+  test-10:
+    <<: *test
```
parkhomenko, line 33: nice, I didn't know about that feature

concourse/pipeline-v5.yml
```diff
@@ -0,0 +1,47 @@
...
+    source:
+      uri: https://github.com/fauna/faunadb-js.git
+      branch: v5
+
+  - name: fauna-js-repository-docs
+    type: git
+    icon: github
+    source:
+      uri: https://github.com/fauna/faunadb-js.git
+      branch: gh-pages-v5
```
parkhomenko, line 15: This can be tricky, not sure we can do it that way,
for jvm driver, we have a separate folder for each version:
we could, probably, do something similar: https://github.com/fauna/faunadb-jvm/tree/gh-pages

parkhomenko (changes requested):
looks good, but I left some comments to consider...

PR #536, opened 2021-09-10
fix: capture stack trace

fireridlle:


milyord (commented):
There's two problems with this fix. One captureStacktTrace is only available in chrome and will throw in other browsers. Two you'll get a stack trace that doesn't originate at the point of throwing the error. If you insist on using this approach you should wrap the call to captureStackTarce in a conditional statement. Also relying on this.constructor can lead to other issues as in some environments the constructor property is not set to configurable: false (granted it is a minor surface for possible future failure).

```js
if (Error.captureStackTrace) {
  Error.captureStackTrace(this, this.constructor)
} 
```

parkhomenko (approved):

PR #537, opened 2021-09-10
fix: ContainsStr FQL to string

fireridlle:
### Notes
[OSS-888](https://faunadb.atlassian.net/browse/OSS-888)

### DIFF
--- a/test/query.test.js
+++ b/test/query.test.js
@@ -1401,6 +1401,9 @@ describe('query', () => {
   })
 
   test('containsstr', () => {
+    expect(fauna.Expr.toString(fauna.query.ContainsStr('ABC', 'A'))).toEqual(
+      'ContainsStr("ABC", "A")'
+    )
     return Promise.all([
       assertQuery(query.ContainsStr('ABC', 'A'), true),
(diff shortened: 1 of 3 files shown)


marrony (changes requested):
Please, add test for this change

Here is a good place https://github.com/fauna/faunadb-js/blob/main/test/values.test.js#L247

parkhomenko (approved):

parkhomenko (dismissed):

test/query.test.js
```diff
@@ -1401,6 +1401,9 @@ describe('query', () => {
   })
 
   test('containsstr', () => {
+    expect(Expr.toString(query.ContainsStr('ABC', 'A'))).toEqual(
```
marrony, line 1404: are we going to move these type of assertions close to the function test? I'm saying because there is a test specific for this case

marrony (approved):
I liked the fact the expression formatting is close to the function assertions, we could adopt this strategy moving forward (and stop adding more test like this to the `values.tests.js`)

marrony (approved):
¯\_(ツ)_/¯

PR #543, opened 2021-10-05
TGB-921 page helper

fireridlle:
### Notes
[OSS-921](https://faunadb.atlassian.net/browse/OSS-921)

### DIFF
--- a/src/PageHelper.js
+++ b/src/PageHelper.js
@@ -193,10 +193,10 @@ PageHelper.prototype._consumePages = function(lambda, reverse) {
   return function(page) {
     var data = []
     page.data.forEach(function(item) {
-      if (item.document) {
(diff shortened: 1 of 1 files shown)


src/PageHelper.js
```diff
@@ -193,10 +193,10 @@ PageHelper.prototype._consumePages = function(lambda, reverse) {
   return function(page) {
     var data = []
     page.data.forEach(function(item) {
```
marrony, line 195: I think this `forEach` block could be removed, it was introduced after we change the api version to 2.7 to keep the backward compatibility but I don't think this was the right with to do. When api versions change and it brings a lot of changes and the users must be aware of this.

fireridlle: done

PR #579, opened 2021-12-07
DOCS-1870: Remove outdated Cloudflare Workers information from JavaScript driver docs

steve-fauna:


concourse/scripts/publish.sh
```diff
@@ -17,7 +17,7 @@ then
   npm run build:package
   echo "Publishing a new version..."
   echo "//registry.npmjs.org/:_authToken=$NPM_TOKEN" > .npmrc
-  npm publish
+  npm publish --tag beta
```
faunaee, line 20: Is this change necessary?

steve-fauna: It was in the `v5` branch that I rebased from. Can one of the other reviewers confirm that it belongs there?

faunaee: If it was in the `v5` branch, it shouldn't be in your PR. You're making a documentation-only change, so the only modifications should be to the documentation.

steve-fauna: OK, change to publish.sh is undone.

faunaee (approved):
LGTM! 👍

PR #582, opened 2021-12-17
Change Client.js `checkNewVersion` default

elersong:
### Notes
[FE-1814]

Currently, the JS driver throws an error if the user is not connected to the internet when testing, which isn't the best user experience. The error is thrown by the FaunaDB client attempting to automatically check for new versions. Updating this to reverse the default version check gives a better user experience.

[FE-1814]: https://faunadb.atlassian.net/browse/FE-1814?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/test/client.test.js
+++ b/test/client.test.js
@@ -408,7 +408,7 @@ describe('Client', () => {
       new Client(util.getCfg())
       await new Promise(resolve => {
         setTimeout(() => {
-          expect(console.info.mock.calls[0][0]).not.toContain(
+          expect(console.info.mock.calls[0]).not.toContain(
             'New faunadb version available'
           )
           resolve()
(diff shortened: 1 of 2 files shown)


test/client.test.js
```diff
@@ -404,25 +404,23 @@ describe('Client', () => {
       resetNotifyAboutNewVersion()
     })
 
-    test('enabled by default', async () => {
+    test('disabled by default', async () => {
```
cleve-fauna, line 407: add a test for enabled if it doesn't already exist.

cleve-fauna (approved):

PR #589, opened 2022-01-19
ENG-3044: Introduce set streaming support

erickpintor:
Changes:

* New event type 'set' added;
* Added type information for set events;
* Test cases.

### DIFF
--- a/src/stream.js
+++ b/src/stream.js
@@ -19,7 +19,7 @@ var http = require('./_http')
 var q = require('./query')
 var util = require('./_util')
 
-var DefaultEvents = ['start', 'error', 'version', 'history_rewrite']
+var DefaultEvents = ['start', 'error', 'version', 'history_rewrite', 'set']
 var DocumentStreamEvents = DefaultEvents.concat(['snapshot'])
 
 /**
@@ -285,13 +285,14 @@ EventDispatcher.prototype.dispatch = function(event) {
 
 /**
  * @typedef {Object} Options
- * @property {string[]} [fields=['action', 'document', 'diff', 'prev']]
+ * @property {string[]} [fields=['action', 'document', 'diff', 'prev', 'index']]
  *   The fields event fields to opt-in during stream subscription. Possible
  *   options:
  *   * 'action': The action type
  *   * 'document': The document's data
  *   * 'diff': The difference between 'document' and 'prev'
  *   * 'prev': The event's previous data
+ *   * 'index': The event's source index, if a set event
  */
 
 /**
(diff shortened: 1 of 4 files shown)


src/stream.js
```diff
@@ -19,7 +19,7 @@ var http = require('./_http')
 var q = require('./query')
 var util = require('./_util')
 
-var DefaultEvents = ['start', 'error', 'version', 'history_rewrite']
+var DefaultEvents = ['start', 'error', 'version', 'history_rewrite', 'set']
```
faunaee, line 22: Later in this file, in lines 286-295, there is a doc block describing the event options, specifically the `fields` options. We need to include coverage of the `index` field.

Also, since client applications might use the available field names, should we define constants for the values? Seems like a miss to have developers use the literal `'index'` to get the `index` field in event objects.

src/types/Client.d.ts
```diff
@@ -6,7 +6,7 @@ import { ExprArg } from './query'
 import RequestResult from './RequestResult'
 import { Subscription, SubscriptionEventHandlers } from './Stream'
 
-type StreamEventFields = 'action' | 'document' | 'diff' | 'prev'
+type StreamEventFields = 'action' | 'document' | 'diff' | 'prev' | 'index'
```
faunaee, line 9: Any way to make the stream event fields available to non-TS devs?

faunaee (changes requested):
Works pretty well. But a couple of small changes could make it better.

src/stream.js
erickpintor: > Later in this file, in lines 286-295, there is a doc block describing the event options, specifically the fields options. We need to include coverage of the index field.

I'll update the JS doc.

> Also, since client applications might use the available field names, should we define constants for the values? Seems like a miss to have developers use the literal 'index' to get the index field in event objects.

I'm unfamiliar with JS library development best practices. I _think_ I'd rather focus on the minimum implementation to enable set-streaming by extending existing behaviour and/or ergonomics while leaving future improvements to next iterations.

src/types/Client.d.ts
erickpintor: They are available as string arguments.

```js
client.stream(.., { fields: ['document', 'index'] })
```

macnealefauna (dismissed):

faunaee (approved):
LGTM! 👍

faunaee:
This PR likely needs to be re-based on `main`, to bring in the changes for version 4.4.2.

faunaee:
Also, the stream tests are failing due to reaching a timeout. It would be great to fix those tests so that they complete successfully.

erickpintor:
This is currently based on top of main. The tests will be failing until the next release of the `fauna/faunadb` docker image.

faunaee (approved):
LGTM! 👍

PR #591, opened 2022-01-21
Fix non-nullable `Select` default parameter type

### LINKED ISSUE #583: Type for `Select` should allow `null` as default value
This is valid FQL: `Select("path", Var("thing"), null)`, but the Typescript type does not allow `null` as a default value.

This was brought up in the community Forums by user, wing: https://forums.fauna.com/t/cannot-specify-null-as-default-value-for-select-in-js-driver-with-typescript/2850

winghouchan:
### Notes

`null` is a valid value in FQL. Fixes #583.

### How to test

1. Pass `null` to the default parameter of [`Select`](https://docs.fauna.com/fauna/current/api/fql/functions/select?lang=javascript). No type error should be raised.

### Screenshots

N/A


ptpaterson:
Internal ticket number is OSS-1033

winghouchan:
Was there a reason for closing this?

doom-weaver:
Apologies, we had to do some branch cleaning and the branch you posted against was removed. We need to reopen this against `v4`.

winghouchan:
Ok, I've reopened in #636 🙂

PR #600, opened 2022-01-26
Chore/dependabot 20220125

rts-rob:
### Notes
[Jira Ticket](https://faunadb.atlassian.net/browse/FE-1889)

Updates all identified severe vulnerabilities.

### How to test

This is a dependency update; test with `npm run test`

### Screenshots

PR #602, opened 2022-01-26
Dependabot fixes 

elersong:
Reset theme for js driver docs to default to remove vulnerable dependency.
Update jest to current version and fix broken Stream tests.
Remove unused `husky` dependency.


package.json
```diff
@@ -27,7 +27,7 @@
     "browserify": "browserify index.js --standalone faunadb -o dist/faunadb.js",
     "browserify-min": "browserify index.js --standalone faunadb | terser -c -m --keep-fnames --keep-classnames -o dist/faunadb-min.js",
     "prettify": "prettier --write \"{src,test}/**/*.{js,ts}\"",
-    "test": "jest --env=node --verbose=true --detectOpenHandles --forceExit ./test",
```
macnealefauna, line 30: What happens when we don't force the exit and detect open handles? Just curious why these went away, not a blocker.

elersong: Nothing happens when we don't force exit. It's not necessary anymore. We fixed the cause of the hangup.
I think we removed the `--detectOpenHandles` because it was flaky, but @rts-rob would have to verify that claim.

rts-rob: Correct - we added `--detectOpenHandles` as a part of #600. No harm in taking it out.

As Grey points out, the force exit was a patch over the actual fix, which this PR includes, and as such is no longer necessary.

faunaee:
Why have the new version notifications been removed? Was the `checkNewVersion` flag insufficient?

If this is a strategic move, are we planning to make the same change in all of the other drivers?

rts-rob:
> Why have the new version notifications been removed? Was the `checkNewVersion` flag insufficient?
> 
> If this is a strategic move, are we planning to make the same change in all of the other drivers?

Removed the notifications as:

1. We had user complaints about the notifications.
2. They require `cross-fetch`, which caused vulnerabilities. This is one step toward removing `cross-fetch`.
3. If we're going to disable it by default by setting it to default, and we don't expect anyone to ever set it, we thought it was better to remove it.

If we know users are actually depending on this for notifications of available updates - which I don't think is idiomatic in JavaScript - we can put it back in with the flag set to `false` by default.

faunaee:
I agree that if we're not going to check for new versions by default, removing the implementation is better. It removes code that would be seldom, if ever, used. If that's what we are going to do here, we should also plan to do it in all of the other drivers.

The removal has doc impact. Is there a Jira issue for the dependabot fixes so that I can link an appropriate doc issue?

rts-rob:
> If that's what we are going to do here, we should also plan to do it in all of the other drivers.

I suggest we remove it in the others as time allows.

> The removal has doc impact. Is there a Jira issue for the dependabot fixes so that I can link an appropriate doc issue?

See [FE-1889](https://faunadb.atlassian.net/browse/FE-1889) for the dependabot issue.

faunaee:
Note that while the new version check has been removed, the flag `checkNewVersion` still exists, so the removal is incomplete.

rts-rob:
I did notice that; I left it in the defaults and the types because it exists in the TypeScript definitions, and I worried removing it could be a breaking change.

Ignoring it doesn't affect _the functionality of the driver_, so should not be considered a major level breaking change.

henryfauna (approved):

package.json
macnealefauna: Sounds good. Go!

PR #608, opened 2022-02-02
ignore `gh-pages` branch in circleci

elersong:
This PR adds configuration to avoid the consistently/uselessly failing builds for jsdoc. Those docs are currently managed via ConcourseCI.


elersong:
This strategy isn't possible with our architecture.

PR #609, opened 2022-02-03
ignore `gh-pages` branch

elersong:
Reconfigure CircleCI to ignore any commits to the `gh-pages` branch. 
The documentation gets published using ConcourseCI, so the attempt to build wastes CI resources and raises useless errors.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -147,31 +147,55 @@ workflows:
       # Stable path
       - core-stable-12:
           context: faunadb-drivers
+          filters:
+            branches:
+              ignore: gh-pages
       - core-stable-14:
           context: faunadb-drivers
           requires:
             - core-stable-12
+          filters:
+            branches:
+              ignore: gh-pages
       - core-stable-16:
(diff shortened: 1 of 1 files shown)


henryfauna (approved):

rts-rob (changes requested):
Per separate conversation - putting the filters here won't keep CircleCI from building on pushes to the `gh-pages` branch, because those pushes don't have _any_ `.circleci/config.yml` associated with them.

Concourse needs to inject the config file into the `gh-pages` branch at build time, before it pushes to this repo. Something like the following should work as a no-op.

```yaml
# No-op workflow for `gh-pages` branch
version: 2.1
​
jobs:
  no-op:
    docker:
      - image: circleci/gh-pages
    steps:
      - checkout
​
workflows:
  no-op:
    jobs:
      - no-op:
          filters:
            branches:
              ignore:
                - gh-pages
```

elersong:
Moving forward with the suggested strategy. Closing this PR

PR #612, opened 2022-02-04
inject CircleCI config into `gh-pages`

elersong:
This PR adds configuration to avoid the consistently/uselessly failing builds for jsdoc. 
Those docs are currently managed via ConcourseCI.

Since the documentation gets regenerated and overwritten on each documentation deploy, the no-op configuration file must be generated and added to the repo before committing.


rts-rob (approved):
LGTM

PR #624, opened 2022-03-20
Propagate Http Errors

zvictor:
Http Errors are not being treated, leading to cryptic `SyntaxError: Unexpected end of JSON input` errors.
This PR raises the http error instead of letting it explode somewhere else in the code.

### DIFF
--- a/src/Client.js
+++ b/src/Client.js
@@ -310,6 +310,10 @@ Client.prototype._execute = function(method, path, data, query, options) {
       })
     )
     .then(function(response) {
+      if(response.status >= 400) {
+        throw new Error(`[HttpError ${response.status}] ${response.body}`)
+      }
+    
       var endTime = Date.now()
       var responseObject = json.parseJSON(response.body)
       var result = new RequestResult(


ptpaterson:
Hi @zvictor and thank you for the PR. Sorry that it's been so long to respond.

This is not an HTTP error, but a JSON parsing error. `SyntaxError: Unexpected end of JSON input` happens when the response is simply not valid JSON and doesn't have anything to do with status codes. We are aware of this issue happening intermittently, for example, when a proxy such as a Vercel function drops part of the payload because they limit the size to 4MB but your query response is bigger than that. Or sending requests to an incorrect endpoint can send HTML responses which also cause problems.

If you have additional scenarios where you get this error, please let us know.

We have an internal issue regarding this, [FE-2895](https://faunadb.atlassian.net/browse/FE-2895).

## Regarding HTTP response codes

The `FaunaHTTPError.raiseForStatusCode` handles cases of http error codes.

https://github.com/fauna/faunadb-js/blob/a4430d6961a0a3ae0f7238620bc85884c9463060/src/Client.js#L353

https://github.com/fauna/faunadb-js/blob/a4430d6961a0a3ae0f7238620bc85884c9463060/src/errors.js#L157-L189

ptpaterson:
I am going to close this PR, but you are welcome to submit a separate solution to the JSON parsing issue. Again, thank you for you feedback!

PR #625, opened 2022-03-24
Add :metrics option to capture headers

elersong:
### Notes
Current limitations of the JS driver prevent easy access to usage metrics provided in the response headers from the DB. This PR adds a `:metrics` option to allow programmatic access to those metrics.

### How to test
`yarn test`

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,6 +154,7 @@ structured as follows:
   } // usage data
 }
 ```
+Metrics returned in the response will be of `Int` data type.
 #### Pagination Helpers
 
 This driver contains helpers to provide a simpler API for consuming paged
--- a/README.md
+++ b/README.md
@@ -127,6 +127,33 @@ createP.then(function(response) {
 `response` is a JSON object containing the FaunaDB response. See the JSDocs for
 `faunadb.Client`.
 
+The `:metrics` option is used during instantiation to create a client that also 
+returns usage information about the queries issued to FaunaDB.
+
+```javascript
+let client = new faunadb.Client({ 
+  secret: 'YOUR_FAUNADB_SECRET',
+  metrics: true 
+})
+```
+
+The `response` object is shaped differently for clients when the `:metrics` option
+is/not set. The default client setting returns the `response` body at root level. 
+When the client is configured to return `:metrics`, the `response` object is 
+structured as follows:
+
+```javascript
+{
+  value: { ... }, // structured response body
+  metrics: {
+    x-compute-ops: XX,
+    x-byte-read-ops: XX,
+    x-byte-write-ops: XX,
+    x-query-time: XX,
+    x-txn-retries: XX
+  } // usage data
+}
+```
 #### Pagination Helpers
 
 This driver contains helpers to provide a simpler API for consuming paged
--- a/src/Client.js
+++ b/src/Client.js
@@ -326,8 +326,15 @@ Client.prototype._execute = function(method, path, data, query, options) {
         endTime
       )
       self._handleRequestResult(response, result, options)
-
-      return responseObject['resource']
+      
+      if (options.withFaunaMetrics) {
+        return { 
+          response: responseObject['resource'],
+          headers: response.headers
+        }
+      } else {
+        return responseObject['resource']
+      }
     })
 }
 
--- a/src/Client.js
+++ b/src/Client.js
@@ -282,6 +282,22 @@ Client.prototype.close = function(opts) {
   return this._http.close(opts)
 }
 
+/**
+ * Executes a query via the FaunaDB Query API.  
+ * See the [docs](https://app.fauna.com/documentation/reference/queryapi),
+ * and the query functions in this documentation.
+ * @param expression {module:query~ExprArg}
+ *   The query to execute. Created from {@link module:query} functions.
+ * @param {?Object} options
+ *   Object that configures the current query, overriding FaunaDB client options.
+ * @param {?string} options.secret FaunaDB secret (see [Reference Documentation](https://app.fauna.com/documentation/intro/security))
+ * @return {external:Promise<Object>} An object containing the FaunaDB response object and the list of query metrics incurred by the request.
+ */
+ Client.prototype.queryWithMetrics = function(expression, options) {
+  options = Object.assign({}, options, { withMetrics: true })
+  return this._execute('POST', '', query.wrap(expression), null, options)
+}
+
 Client.prototype._execute = function(method, path, data, query, options) {
   query = util.defaults(query, null)
 
--- a/src/Client.js
+++ b/src/Client.js
@@ -356,7 +356,7 @@ Client.prototype._execute = function(method, path, data, query, options) {
         'x-txn-retries'
       ]
 
-      if (options?.metrics) {
+      if (options && options.metrics) {
         return { 
           value: responseObject['resource'],
           metrics:  Object.fromEntries(
--- a/src/Client.js
+++ b/src/Client.js
@@ -165,6 +165,8 @@ var values = require('./values')
  *   which has the highest priority and overrides the option passed into the Client constructor.
  * @param {?boolean} options.checkNewVersion
  *   Enabled by default. Prints a message to the terminal when a newer driver is available.
+ * @param {?boolean} options.metrics
+ *   Disabled by default. Controls whether or not query metrics are returned.
  */
 function Client(options) {
   var http2SessionIdleTime = getHttp2SessionIdleTime()
--- a/src/Client.js
+++ b/src/Client.js
@@ -346,10 +346,10 @@ Client.prototype._execute = function(method, path, data, query, options) {
       self._handleRequestResult(response, result, options)
 
       const metricsHeaders = [
-        'x-query-time',
-        'x-read-ops',
-        'x-write-ops',
         'x-compute-ops',
+        'x-byte-read-ops',
+        'x-byte-write-ops',
+        'x-query-time',
         'x-txn-retries',
       ]
 
--- a/src/Client.js
+++ b/src/Client.js
@@ -361,11 +361,9 @@ Client.prototype._execute = function(method, path, data, query, options) {
       if (options && options.metrics) {
         return {
           value: responseObject['resource'],
-          metrics: Object.fromEntries(
-            Object.entries(response.headers).filter(([key]) =>
-              metricsHeaders.includes(key)
-            )
-          ),
+          metrics: Object.fromEntries(Array.from(Object.entries(response.headers)).
+            map(([ k,v ]) => [k, parseInt(v)]).
+            filter( ([k,v]) => metricsHeaders.includes(k) )),
         }
       } else {
         return responseObject['resource']
--- a/src/Client.js
+++ b/src/Client.js
@@ -291,7 +291,7 @@ Client.prototype.close = function(opts) {
  * @param {?Object} options
  *   Object that configures the current query, overriding FaunaDB client options.
  * @param {?string} options.secret FaunaDB secret (see [Reference Documentation](https://app.fauna.com/documentation/intro/security))
- * @return {external:Promise<Object>} An object containing the FaunaDB response object and the list of query metrics incurred by the request.
+ * @return {external:Promise<Object>} {value, metrics} An object containing the FaunaDB response object and the list of query metrics incurred by the request.
  */
  Client.prototype.queryWithMetrics = function(expression, options) {
   options = Object.assign({}, options, { metrics: true })
--- a/src/Client.js
+++ b/src/Client.js
@@ -294,7 +294,7 @@ Client.prototype.close = function(opts) {
  * @return {external:Promise<Object>} An object containing the FaunaDB response object and the list of query metrics incurred by the request.
  */
  Client.prototype.queryWithMetrics = function(expression, options) {
-  options = Object.assign({}, options, { withMetrics: true })
+  options = Object.assign({}, options, { metrics: true })
   return this._execute('POST', '', query.wrap(expression), null, options)
 }
 
--- a/src/Client.js
+++ b/src/Client.js
@@ -344,15 +344,11 @@ Client.prototype._execute = function(method, path, data, query, options) {
       )
       self._handleRequestResult(response, result, options)
       
-      const metricsHeaders = [ 
-        'x-query-bytes-in', 
-        'x-query-bytes-out', 
-        'x-query-time', 
+      const metricsHeaders = [
+        'x-compute-ops',
         'x-read-ops', 
         'x-write-ops', 
-        'x-compute-ops', 
-        'x-storage-bytes-read', 
-        'x-storage-bytes-write', 
+        'x-query-time', 
         'x-txn-retries'
       ]
 
--- a/src/Client.js
+++ b/src/Client.js
@@ -192,6 +192,7 @@ function Client(options) {
   this._observer = options.observer
   this._http = new http.HttpClient(options)
   this.stream = stream.StreamAPI(this)
+  this.metrics = options.metrics
 }
 
 /**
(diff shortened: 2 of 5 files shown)


src/_util.js
```diff
@@ -330,7 +330,7 @@ function getNodeRuntimeEnv() {
  * @param {any} def The default value.
  * @private
  */
-function defaults(obj, def) {
+function defaults(obj, def = { withFaunaMetrics: false }) {
```
doom-weaver: changing this code would change the behavior of all users of this utility function.


So instead of bolting this in here how about you just use the defaults method inside _execute.

Like:

```
options = util.defaults(options, {withFaunaMetrics: false})
```
this would be the same pattern as query is doing here:

https://github.com/fauna/faunadb-js/blob/ee6f53f9c985bde41976743530e3846dee058587/src/Client.js#L285

src/Client.js
```diff
@@ -326,8 +326,15 @@ Client.prototype._execute = function(method, path, data, query, options) {
         endTime
       )
       self._handleRequestResult(response, result, options)
-
-      return responseObject['resource']
+      
+      if (options.withFaunaMetrics) {
+        return { 
+          response: responseObject['resource'],
+          headers: response.headers
```
faunaee: headers !== metrics. Sure, metrics are part of the headers, but not all of the response headers are metrics.

We should return `metrics`, with the subset of the headers that represent metrics.

The headers that represent metrics are:

```
[
  'x-query-bytes-in',
  'x-query-bytes-out',
  'x-query-time',
  'x-read-ops',
  'x-write-ops',
  'x-compute-ops',
  'x-storage-bytes-read',
  'x-storage-bytes-write',
  'x-txn-retries',
]
```

src/Client.js
```diff
@@ -326,8 +326,15 @@ Client.prototype._execute = function(method, path, data, query, options) {
         endTime
       )
       self._handleRequestResult(response, result, options)
-
-      return responseObject['resource']
+      
+      if (options.withFaunaMetrics) {
```
faunaee:
```suggestion
      if (options.metrics) {
```

This is `fauna-shell`, so the `Fauna` is redundant. And it would be much more succinct to just call:

```
  secret: "secret",
  metrics: true
})
```

src/Client.js
faunaee: Is there a test that exercises the metrics return? If not, shouldn't there be at least one?

src/Client.js
faunaee: The `README` file should be updated to document this option.

faunaee (dismissed):
Nice! Many JS driver users will appreciate this feature. I made a few suggestions.

ptpaterson:
Can we add another method to Client as a shorthand for setting the `withMetrics` option?  This would harmonize it with the JVM drivers.  Something like:

```javascript
/**
 * Executes a query via the FaunaDB Query API.  
 * See the [docs](https://app.fauna.com/documentation/reference/queryapi),
 * and the query functions in this documentation.
 * @param expression {module:query~ExprArg}
 *   The query to execute. Created from {@link module:query} functions.
 * @param {?Object} options
 *   Object that configures the current query, overriding FaunaDB client options.
 * @param {?string} options.secret FaunaDB secret (see [Reference Documentation](https://app.fauna.com/documentation/intro/security))
 * @return {external:Promise<Object>} An object containing the FaunaDB response object and the list of query metrics incurred by the request.
 */
Client.prototype.queryWithMetrics = function(expression, options) {
  options = Object.assign({}, options, { withMetrics: true })
  return this._execute('POST', '', query.wrap(expression), null, options)
}
```

src/Client.js
ptpaterson: The JVM client's [MetricsResponse class](https://github.com/fauna/faunadb-jvm/blob/5ffdba7c0ffe8e642c7c2ed2131d41d677de8037/faunadb-java/src/main/java/com/faunadb/client/types/MetricsResponse.java#L13-L15) has `metricsMap` and `value` fields.  The "Map" part is redundant in JS, so if we want to harmonize with JVM driver, then we could return 

```javascript
{
  value:  responseObject['resource'],
  metrics: getMetrics(response.headers)
}
```

Also highlighting that headers `!==` metrics

src/_util.js
elersong: It looks like the client default values are actually here https://github.com/fauna/faunadb-js/pull/625/files/db7057a186f80944e548476736b361dc613beb66#diff-b880f47f381d68dcd7473fb094acb874cd50b1140cdf2bfd7a368fa509738073L172

elersong: I updated them to include the `metrics` default

src/Client.js
elersong: Done

src/Client.js
```diff
@@ -281,6 +282,22 @@ Client.prototype.close = function(opts) {
...
+ * and the query functions in this documentation.
+ * @param expression {module:query~ExprArg}
+ *   The query to execute. Created from {@link module:query} functions.
+ * @param {?Object} options
+ *   Object that configures the current query, overriding FaunaDB client options.
+ * @param {?string} options.secret FaunaDB secret (see [Reference Documentation](https://app.fauna.com/documentation/intro/security))
+ * @return {external:Promise<Object>} An object containing the FaunaDB response object and the list of query metrics incurred by the request.
+ */
+ Client.prototype.queryWithMetrics = function(expression, options) {
+  options = Object.assign({}, options, { withMetrics: true })
```
doom-weaver: this is using `withMetrics` but elsewhere `metrics` is used.

```suggestion
  options = Object.assign({}, options, { metrics: true })
```

src/Client.js
```diff
@@ -326,8 +343,29 @@ Client.prototype._execute = function(method, path, data, query, options) {
...
+      const metricsHeaders = [ 
+        'x-query-bytes-in', 
+        'x-query-bytes-out', 
+        'x-query-time', 
+        'x-read-ops', 
+        'x-write-ops', 
+        'x-compute-ops', 
+        'x-storage-bytes-read', 
+        'x-storage-bytes-write', 
+        'x-txn-retries'
```
doom-weaver: i think you also need `x-query-byte-read-ops` and `x-query-byte-write-ops`. See [docs](https://docs.fauna.com/fauna/current/learn/understanding/billing#perquery).

When I query via extending the current driver I get teh following headers:

```
responseHeaders: {
      ':status': 200,
      'x-txn-time': '1648432840766000',
      'x-compute-ops': '1',
      'x-read-ops': '1',
      'x-byte-read-ops': '16',
      'x-byte-write-ops': '0',
      'x-write-ops': '0',
      'x-query-time': '216',
      'x-query-bytes-in': '31',
      'x-query-bytes-out': '24',
      'x-storage-bytes-read': '200',
      'x-storage-bytes-write': '0',
      'x-txn-retries': '0',
      'x-faunadb-build': '22.03.01-39f78e0',
      'content-length': '24',
      'content-type': 'application/json;charset=utf-8'
}
 ```

From these the metrics ones are:

```
 {
      'x-compute-ops': '1',
      'x-read-ops': '1',
      'x-byte-read-ops': '16',
      'x-byte-write-ops': '0',
      'x-write-ops': '0',
      'x-query-time': '216',
      'x-query-bytes-in': '31',
      'x-query-bytes-out': '24',
      'x-storage-bytes-read': '200',
      'x-storage-bytes-write': '0',
      'x-txn-retries': '0'
}
```

I'm curious if others think we should even include  the `x-read-ops, x-query-bytes-in, x-query-bytes-out, x-write-ops, and x-storage-bytes-write` headers since they  "refer to legacy query accounting, and are not used in billing calculations." Including might just lead to confusion.

As Ewan called out below `x-storage-bytes-read` is also deprecated.

Without those we'd have:

```
 {
      'x-compute-ops': '1',
      'x-byte-read-ops': '16',
      'x-byte-write-ops': '0',
      'x-query-time': '216',
      'x-txn-retries': '0',
}
```

@faunaee @ptpaterson @rts-rob

**Review by doom-weaver — CHANGES_REQUESTED** [review]

File: src/Client.js (hunk 5 above)
Comment (rts-rob) [reply src/Client.js]: Since we're introducing new functionality I propose we limit the information provided to only what's useful and used.

@cleve-fauna You didn't explicitly mention `x-storage-bytes-read` - is it legacy as well?

Comment (doom-weaver) [reply src/Client.js]: my bad its not called out as legacy; added it to the list.

Please double check my reading here: https://docs.fauna.com/fauna/current/learn/understanding/billing#perquery 😆

Comment (faunaee) [reply src/Client.js]: I'm good with removing the legacy headers from the `metrics` return value. If users still need to access those, they can continue to provide an observer callback.

Comment (doom-weaver) [reply src/Client.js]: Its a little vague on `x-storage-bytes-read`  - we don't use this in any billing. Do you see any reason to keep it @faunaee ?

Comment (faunaee) [reply src/Client.js]: No. `x-storage-bytes-read` and `x-storage-bytes-write` have been superceded by `x-bytes-read-ops` and `x-bytes-write-ops`, respectively.

That `x-storage-bytes-read` isn't listed as "legacy" in the docs is a bug. I'll get that fixed via DOCS-2011.

Comment (elersong) [reply src/Client.js]: I've updated to remove the following headers from the `:metrics`

```
'x-query-bytes-in', 
'x-query-bytes-out', 
'x-compute-ops', 
'x-storage-bytes-read', 
'x-storage-bytes-write', 
```

File: src/Client.js [hunk 6]
```diff
@@ -326,8 +343,25 @@ Client.prototype._execute = function(method, path, data, query, options) {
...
+        'x-read-ops', 
+        'x-write-ops', 
+        'x-query-time', 
+        'x-txn-retries'
+      ]
 
-      return responseObject['resource']
+      if (options && options.metrics) {
+        return { 
+          value: responseObject['resource'],
```
faunaee, line 363: Is there a benefit to returning an object here? What about this usage:

```
const [ response, metrics ] = client.queryWithMetrics(q.Now())
```

If there is value to returning an object, can we call it `response` instead of `value`?

src/Client.js
rts-rob: Is there a mismatch here? I understood @cleve-fauna 's comment above to mean we should _remove_:

```js
'x-read-ops',
'x-query-bytes-in',
'x-query-bytes-out',
'x-write-ops',
'x-storage-bytes-read',
'x-storage-bytes-write',
```

_If_ that's correct, then after fde473b we need a new commit that adds back in:
```js
'x-compute-ops',
```
and removes:
```js
'x-read-ops',
'x-write-ops',
```

src/Client.js
```diff
@@ -281,6 +282,22 @@ Client.prototype.close = function(opts) {
...
+/**
+ * Executes a query via the FaunaDB Query API.  
+ * See the [docs](https://app.fauna.com/documentation/reference/queryapi),
+ * and the query functions in this documentation.
+ * @param expression {module:query~ExprArg}
+ *   The query to execute. Created from {@link module:query} functions.
+ * @param {?Object} options
+ *   Object that configures the current query, overriding FaunaDB client options.
+ * @param {?string} options.secret FaunaDB secret (see [Reference Documentation](https://app.fauna.com/documentation/intro/security))
+ * @return {external:Promise<Object>} An object containing the FaunaDB response object and the list of query metrics incurred by the request.
```
faunaee: If we return an object, should we note the names of the fields here?

faunaee (dismissed):
A couple of minor suggestions.

Also, we should include a test and an update to `README.md` describing this feature.

src/Client.js
elersong: `value` was a request to align the JS driver with the JVM driver, which already has a `queryWithMetrics` workflow.

test/client.test.js
```diff
@@ -386,12 +405,11 @@ describe('Client', () => {
...
-
 })
 
-function assertHeader(headers, name) {
-  expect(headers[name]).not.toBeNull()
-  expect(parseInt(headers[name])).toBeGreaterThanOrEqual(0)
+function assertMetric(metrics, name) {
+  expect(metrics[name]).not.toBeNull()
+  expect(parseInt(metrics[name])).toBeGreaterThanOrEqual(0)
 }
```
doom-weaver: I notice here you have to convert the metric to an int.

It would be a better UX if we did this in the driver code itself - otherwise every client is going to have to do this conversion.

src/Client.js
```diff
@@ -327,7 +350,26 @@ Client.prototype._execute = function(method, path, data, query, options) {
...
+        'x-query-time',
+        'x-txn-retries',
+      ]
+
+      if (options && options.metrics) {
+        return {
+          value: responseObject['resource'],
+          metrics: Object.fromEntries(
+            Object.entries(response.headers).filter(([key]) =>
+              metricsHeaders.includes(key)
```
doom-weaver: converting the metrics to an int ourselves here we eliminate forcing that logic on each client.

doom-weaver (changes requested):

README.md
```diff
@@ -127,6 +127,33 @@ createP.then(function(response) {
...
+structured as follows:
+
+```javascript
+{
+  value: { ... }, // structured response body
+  metrics: {
+    x-compute-ops: XX,
+    x-byte-read-ops: XX,
+    x-byte-write-ops: XX,
+    x-query-time: XX,
```
doom-weaver, line 152: if we change the metrics to be an object pointing to integer values let's call it out in the doc as well so that users don't have to investigate the type we return.

test/client.test.js
elersong: Good call

src/Client.js
elersong: done

src/Client.js
```diff
@@ -327,7 +350,24 @@ Client.prototype._execute = function(method, path, data, query, options) {
...
+        'x-query-time',
+        'x-txn-retries',
+      ]
+
+      if (options && options.metrics) {
+        return {
+          value: responseObject['resource'],
+          metrics: Object.fromEntries(Array.from(Object.entries(response.headers)).
+            map(([ k,v ]) => [k, parseInt(v)]).
+            filter( ([k,v]) => metricsHeaders.includes(k) )),
```
doom-weaver: nit - filter and then map, save some computation
```suggestion
            filter( ([k,v]) => metricsHeaders.includes(k) ).
            map(([ k,v ]) => [k, parseInt(v)]))
            ,
```

test/client.test.js
```diff
@@ -40,6 +40,25 @@ describe('Client', () => {
...
+    assertMetric(response.metrics, 'x-query-time')
+    assertMetric(response.metrics, 'x-txn-retries')
+  })
+
+  test('returns query metrics if using the queryWithMetrics function', async () => {
+    const response = await client.queryWithMetrics(query.Add(1, 1))
+    assertMetric(response.metrics, 'x-compute-ops')
+    assertMetric(response.metrics, 'x-byte-read-ops')
+    assertMetric(response.metrics, 'x-byte-write-ops')
+    assertMetric(response.metrics, 'x-query-time')
```
doom-weaver, line 58: please add an assert on the value part of the result too.

src/Client.js
```diff
@@ -327,7 +350,25 @@ Client.prototype._execute = function(method, path, data, query, options) {
...
+        'x-query-time',
+        'x-txn-retries',
+      ]
+
+      if (options && options.metrics) {
+        return {
+          value: responseObject['resource'],
+          metrics: Object.fromEntries(Array.from(Object.entries(response.headers)).
+            filter( ([k,v]) => metricsHeaders.includes(k) )).
+            map(([ k,v ]) => [k, parseInt(v)])
```
doom-weaver:
```suggestion
          metrics: Object.fromEntries(
            Array.from(Object.entries(response.headers)).
              filter( ([k,v]) => metricsHeaders.includes(k) ).
              map( ([ k,v ]) => [k, parseInt(v)])
            )
```

doom-weaver (approved):

PR #628, opened 2022-03-29
Add types for Client.queryWithMetrics

ptpaterson:
#625 did not add types for the Client class.

### DIFF
--- a/src/types/Client.d.ts
+++ b/src/types/Client.d.ts
@@ -49,9 +49,24 @@ interface StreamApi
   document: StreamFn<SubscriptionEventHandlers>
 }
 
+interface MetricsResponse<T = object> {
+  value: T
+  metrics: {
+    'x-compute-ops': number
+    'x-byte-read-ops': number
+    'x-byte-write-ops': number
+    'x-query-time': number
+    'x-txn-retries': number
+  }
+}
+
 export default class Client {
   constructor(opts?: ClientConfig)
   query<T = object>(expr: ExprArg, options?: QueryOptions): Promise<T>
+  queryWithMetrics<T = object>(
+    expr: ExprArg,
+    options?: QueryOptions
+  ): Promise<MetricsResponse<T>>
   paginate(expr: Expr, params?: object, options?: QueryOptions): PageHelper
   ping(scope?: string, timeout?: number): Promise<string>
   close(opts?: { force?: boolean }): Promise<void>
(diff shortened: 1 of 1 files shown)


src/types/Client.d.ts
```diff
@@ -49,9 +49,24 @@ interface StreamApi
...
+  }
+}
+
 export default class Client {
   constructor(opts?: ClientConfig)
   query<T = object>(expr: ExprArg, options?: QueryOptions): Promise<T>
+  queryWithMetrics<T = object>(
+    expr: ExprArg,
+    options?: QueryOptions
+  ): Promise<MetricsResponse<T>>
```
doom-weaver: query now returns an or type.

: `Promise<T> | Promise<MetricsResponse<T>>`

doom-weaver (dismissed):

src/types/Client.d.ts
```diff
@@ -49,9 +49,24 @@ interface StreamApi
...
+    'x-byte-read-ops': number
+    'x-byte-write-ops': number
+    'x-query-time': number
+    'x-txn-retries': number
+  }
+}
+
 export default class Client {
   constructor(opts?: ClientConfig)
   query<T = object>(expr: ExprArg, options?: QueryOptions): Promise<T>
```
doom-weaver:
```suggestion
  query<T = object>(expr: ExprArg, options?: QueryOptions): Promise<T> | Promise<MetricsResponse<T>>
```

We need to use a Union type here for the case where the user explicitly passes the `metrics: true` option

doom-weaver (approved):

src/types/Client.d.ts
```diff
@@ -49,9 +49,24 @@ interface StreamApi
...
+    'x-byte-write-ops': number
+    'x-query-time': number
+    'x-txn-retries': number
+  }
+}
+
 export default class Client {
   constructor(opts?: ClientConfig)
-  query<T = object>(expr: ExprArg, options?: QueryOptions): Promise<T>
+  query<T = object>(expr: ExprArg, options?: QueryOptions): Promise<T> | Promise<MetricsResponse<T>>
```
Akryum, line 65: This is a breaking change for TypeScript projects :(

Akryum: Example:

image

```
error TS2322: Type 'boolean | MetricsResponse<boolean>' is not assignable to type 'boolean'.
```

PR #633, opened 2022-03-30
Fix backward compatibility bug introduced in 4.5.3 that causes query to return a differnt type

doom-weaver:
### Notes
4.5.3 introduced a bug that broke backward compatibility - it started returning a union of two types on `query` instead of just a single type.

This commit fixes that bug by reverting query to the old interface. It does so by:

- disabling the ability of a user to configure a client such that it always returns metrics on query
- disabling the ability to configure a query call with options to return metrics

Customers will instead need to use queryWithMetrics to get these metrics.

### Testing

Pending circleci

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 4.5.4
+- Disable ability to configure a client and the query method from returning metrics when calling query - fixing bug introduced in 4.5.3 that breaks backward compatibility. Continue supporting queryWithMetrics. [#]().
+
 ## 4.5.3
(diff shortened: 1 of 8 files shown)


faunaee (dismissed):
Only performed a visual inspection, but LGTM! 👍

henryfauna (approved):

grempe:
Thanks for the revert. This caused some pain and some hours.

doom-weaver:
> Thanks for the revert. This caused some pain and some hours.

Sorry for the bad release @grempe! Happy to hear you are unstuck.

PR #642, opened 2022-06-08
Set max session idle time

henryfauna:
### Notes
[Jira Ticket](https://faunadb.atlassian.net/browse/FE-2238)

Previously "Infinity" was a possible value for `http2SessionIdleTime`. This change enforces a maximum value of `5000` milliseconds.

### How to test
`yarn test`

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ This Driver supports and is tested on:
 - Node.js [*Current*, *Active LTS*, and *Maintenance LTS* releases](https://nodejs.org/en/about/releases/)
   - *Current* - v17
   - *Active LTS* - currently v16
-  - *Maintenance LTS* - v12 and v14
+  - *Maintenance LTS* - v12 and v14 
 - Chrome
 - Firefox
 - Safari
--- a/README.md
+++ b/README.md
@@ -19,10 +19,10 @@ reference](https://docs.fauna.com/fauna/current/api/fql/).
 
 This Driver supports and is tested on:
 
-- Node.js [*Current*, *Active LTS*, and *Maintenance LTS* releases](https://nodejs.org/en/about/releases/)
-  - *Current* - v17
-  - *Active LTS* - currently v16
-  - *Maintenance LTS* - v12 and v14 
+- Node.js [_Current_, _Active LTS_, and _Maintenance LTS_ releases](https://nodejs.org/en/about/releases/)
+  - _Current_ - v17
+  - _Active LTS_ - currently v16
+  - _Maintenance LTS_ - v12 and v14
 - Chrome
 - Firefox
 - Safari
@@ -127,13 +127,13 @@ createP.then(function(response) {
 `response` is a JSON object containing the FaunaDB response. See the JSDocs for
 `faunadb.Client`.
 
-The `metrics` option is used during instantiation to create a client that also 
+The `metrics` option is used during instantiation to create a client that also
 returns usage information about the queries issued to FaunaDB.
 
 ```javascript
-let client = new faunadb.Client({ 
+let client = new faunadb.Client({
   secret: 'YOUR_FAUNADB_SECRET',
-  metrics: true 
+  metrics: true,
 })
 ```
 
@@ -155,7 +155,9 @@ time, and transaction retires consumed by your query:
   } // usage data
 }
 ```
+
 Metrics returned in the response will be of `number` data type.
+
 #### Pagination Helpers
 
 This driver contains helpers to provide a simpler API for consuming paged
@@ -277,17 +279,17 @@ const client = new faunadb.Client({
 When running on the Node.js platform, the Fauna client uses [HTTP/2 multiplexing](https://stackoverflow.com/questions/36517829/what-does-multiplexing-mean-in-http-2)
 to reuse the same session for many simultaneous requests. After all open requests
 have been resolved, the client will keep the session open for a period of time
-(500ms by default) to be reused for any new requests.
+to be reused for any new requests.
 
 The `http2SessionIdleTime` parameter may be used to control how long the HTTP/2
 session remains open while the connection is idle. To save on the overhead of
-closing and re-opening the session, set `http2SessionIdleTime` to a longer time
---- or even `Infinity`, to keep the session alive indefinitely.
+closing and re-opening the session, set `http2SessionIdleTime` to a longer time.
+The default value is 500ms and the maximum value is 5000ms.
 
 While an HTTP/2 session is alive, the client will hold the Node.js event loop
 open; this prevents the process from terminating. Call `Client#close` to manually
 close the session and allow the process to terminate. This is particularly
-important if `http2SessionIdleTime` is long or `Infinity`:
+important if `http2SessionIdleTime` is long:
 
 ```javascript
 // sample.js (run it with "node sample.js" command)
@@ -296,8 +298,8 @@ const { Client, query: Q } = require('faunadb')
 async function main() {
   const client = new Client({
     secret: 'YOUR_FAUNADB_SECRET',
-    http2SessionIdleTime: Infinity,
-    //                    ^^^ Infinity or non-negative integer
+    http2SessionIdleTime: 1000,
+    //                    ^^^ Non-negative integer
   })
   const output = await client.query(Q.Add(1, 1))
 
@@ -309,7 +311,6 @@ async function main() {
 
 main().catch(console.error)
 ```
-
 
 ## Known issues
 
--- a/README.md
+++ b/README.md
@@ -282,11 +282,14 @@ have been resolved, the client will keep the session open for a period of time
 (500ms by default) to be reused for any new requests.
 
 The `http2SessionIdleTime` parameter may be used to control how long the HTTP/2
-session remains open while the connection is idle. To save on the overhead of
+session remains open while the query connection is idle. To save on the overhead of
 closing and re-opening the session, set `http2SessionIdleTime` to a longer time.
 The default value is 500ms and the maximum value is 5000ms.
 
-While an HTTP/2 session is alive, the client will hold the Node.js event loop
+Note that `http2SessionIdleTime` has no effect on a stream connection: a stream
+is a long-lived connection that is intended to be held open indefinitely.
+
+While an HTTP/2 session is alive, the client holds the Node.js event loop
 open; this prevents the process from terminating. Call `Client#close` to manually
 close the session and allow the process to terminate. This is particularly
 important if `http2SessionIdleTime` is long:
@@ -298,15 +301,13 @@ const { Client, query: Q } = require('faunadb')
 async function main() {
   const client = new Client({
     secret: 'YOUR_FAUNADB_SECRET',
-    http2SessionIdleTime: 1000,
-    //                    ^^^ Non-negative integer
+    http2SessionIdleTime: 1000, // Must be a non-negative integer
   })
   const output = await client.query(Q.Add(1, 1))
 
   console.log(output)
 
   client.close()
-  //     ^^^ If it's not called then the process won't terminate
 }
 
 main().catch(console.error)
--- a/README.md
+++ b/README.md
@@ -19,10 +19,10 @@ reference](https://docs.fauna.com/fauna/current/api/fql/).
 
 This Driver supports and is tested on:
 
-- Node.js [_Current_, _Active LTS_, and _Maintenance LTS_ releases](https://nodejs.org/en/about/releases/)
-  - _Current_ - v17
-  - _Active LTS_ - currently v16
-  - _Maintenance LTS_ - v12 and v14
+- Node.js [*Current*, *Active LTS*, and *Maintenance LTS* releases](https://nodejs.org/en/about/releases/)
+  - *Current* - v17
+  - *Active LTS* - currently v16
+  - *Maintenance LTS* - v12 and v14
 - Chrome
 - Firefox
 - Safari
@@ -127,13 +127,13 @@ createP.then(function(response) {
 `response` is a JSON object containing the FaunaDB response. See the JSDocs for
 `faunadb.Client`.
 
-The `metrics` option is used during instantiation to create a client that also
+The `metrics` option is used during instantiation to create a client that also 
 returns usage information about the queries issued to FaunaDB.
 
 ```javascript
-let client = new faunadb.Client({
+let client = new faunadb.Client({ 
   secret: 'YOUR_FAUNADB_SECRET',
-  metrics: true,
+  metrics: true 
 })
 ```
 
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -277,17 +279,17 @@ const client = new faunadb.Client({
 When running on the Node.js platform, the Fauna client uses [HTTP/2 multiplexing](https://stackoverflow.com/questions/36517829/what-does-multiplexing-mean-in-http-2)
 to reuse the same session for many simultaneous requests. After all open requests
 have been resolved, the client will keep the session open for a period of time
-(500ms by default) to be reused for any new requests.
+to be reused for any new requests.
```
doom-weaver: please leave the 500 ms in - that is still valid.

README.md
```diff
@@ -19,10 +19,10 @@ reference](https://docs.fauna.com/fauna/current/api/fql/).
...
 This Driver supports and is tested on:
 
-- Node.js [*Current*, *Active LTS*, and *Maintenance LTS* releases](https://nodejs.org/en/about/releases/)
-  - *Current* - v17
-  - *Active LTS* - currently v16
-  - *Maintenance LTS* - v12 and v14 
+- Node.js [_Current_, _Active LTS_, and _Maintenance LTS_ releases](https://nodejs.org/en/about/releases/)
+  - _Current_ - v17
+  - _Active LTS_ - currently v16
+  - _Maintenance LTS_ - v12 and v14
```
doom-weaver: why the formatting changes?

src/Client.js
```diff
@@ -299,7 +296,14 @@ Client.prototype.queryWithMetrics = function(expression, options) {
...
 
-Client.prototype._execute = function(method, path, data, query, options, returnMetrics = false) {
+Client.prototype._execute = function(
+  method,
+  path,
+  data,
+  query,
+  options,
+  returnMetrics = false
+) {
```
doom-weaver: let's avoid unnecessary formatting changes. There are several other places with such formatting changes in this file.

test/client.test.js
```diff
@@ -412,6 +412,56 @@ describe('Client', () => {
       requiredKeys.every(key => driverEnvHeader.includes(key))
     ).toBeDefined()
   })
+
+  test('http2SessionIdleTime env overrides client config', async () => {
```
doom-weaver, line 421: - add a test for passing garbage into the env var - something that is not `Infinity` and not an integer.
- add a test for pssing garbage in the options
- add a test for passing a negative number in the env var
- add a test for passing a negative number in the options

src/Client.js
```diff
@@ -182,13 +179,13 @@ function Client(options) {
...
+    http2SessionIdleTime: 500,
     checkNewVersion: false,
   })
 
-  if (http2SessionIdleTime.shouldOverride) {
-    options.http2SessionIdleTime = http2SessionIdleTime.value
-  }
+  options.http2SessionIdleTime = getHttp2SessionIdleTime(
+    options.http2SessionIdleTime
+  )
```
doom-weaver: how about we keep the call to `getHttp2SessionIdleTimeout` on on line 172 and just plug it in on line 182 - skip this whole branching structure on line 186.

src/Client.js
```diff
@@ -384,17 +390,32 @@ Client.prototype._handleRequestResult = function(response, result, options) {
...
 
-  return {
-    shouldOverride: useEnvVar,
-    value: useEnvVar ? parsed : 500,
-  }
+  const envIdleTime = util.getEnvVariable('FAUNADB_HTTP2_SESSION_IDLE_TIME')
+
+  const parsedEnv = parseInt(envIdleTime, 10)
+  const parsedConfig = parseInt(configuredIdleTime, 10)
+  var value = parsedEnv || parsedConfig || defaultIdleTime
```
doom-weaver: you need to handle the case where parsedEnv and parsedConfig are negative.

src/Client.js
```diff
@@ -384,17 +390,32 @@ Client.prototype._handleRequestResult = function(response, result, options) {
...
+
+  const isGreaterThanMax = value > maxIdleTime
+  const isInfinity =
+    envIdleTime === 'Infinity' || configuredIdleTime === 'Infinity'
+
+  if (isGreaterThanMax)
+    console.warn(
+      `The value set for http2SessionIdleTime exceeds the maximum value of ${maxIdleTime} milliseconds.`
+    )
```
doom-weaver: this is vague. Make it clear you will ceiling the `http2SessionIdleTime` at 5 seconds. The isInfinity block is vague as well.

I suspect you can just use one block.

If you set `isGreaterThanMax = value > maxIldeTime || isInfinity` then you could just use one block.

README.md
henryfauna: VS code auto formatter started doing stuff. Should be fixed now

src/Client.js
henryfauna: If we plug it in on line 182 it can be overwitten as that function is just stuffing in defaults if things haven't been explicitly set. I agree though the way I'm doing it now is a little weird. I'll see if I can come up with something better.

test/client.test.js
henryfauna: These should all be covered now.

src/Client.js
```diff
@@ -384,17 +383,37 @@ Client.prototype._handleRequestResult = function(response, result, options) {
...
+  const defaultIdleTime = 500
+  const envIdleTime = util.getEnvVariable('FAUNADB_HTTP2_SESSION_IDLE_TIME')
 
-  return {
-    shouldOverride: useEnvVar,
-    value: useEnvVar ? parsed : 500,
+  var value = defaultIdleTime
+  // attemp to set the idle time to the env value and then the configured value
+  const values = [envIdleTime, configuredIdleTime]
+  for (let i = 0; i <= values.length; i++) {
```
doom-weaver: you're going to overrun the bounds of the array:

```suggestion
  const values = [envIdleTime, configuredIdleTime]
  for (let i = 0; i < values.length; i++) {
```

for future work you can also consider using `for...of` for iterables:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of

src/Client.js
```diff
@@ -158,8 +158,7 @@ var values = require('./values')
  *   Sets the maximum amount of time (in milliseconds) for query execution on the server
  * @param {?number} options.http2SessionIdleTime
  *   Sets the maximum amount of time (in milliseconds) an HTTP2 session may live
- *   when there's no activity. Must either be a non-negative integer, or Infinity to allow the
- *   HTTP2 session to live indefinitely (use `Client#close` to manually terminate the client).
+ *   when there's no activity. Must either be a non-negative integer.
```
doom-weaver: Clarify the documentation here.

```suggestion
 *   when there's no activity. Must be a non-negative integer, with a maximum value of 5000.
 *   If an invalid value is passed a default of 500 ms will be applied. If a value exceeding 5000 ms
 *  is passed (e.g. Infinity) the maximum of 5000 ms will be applied.
```

src/Client.js
```diff
@@ -384,17 +383,37 @@ Client.prototype._handleRequestResult = function(response, result, options) {
...
+    if (isNegative) continue
+    if (!isInfinity && !parsedValue) continue
+    // if we did get something valid constrain it to the ceiling
+    value = parsedValue
+    if (isGreaterThanMax) {
+      value = maxIdleTime
+      console.warn(
+        `The value set for http2SessionIdleTime exceeds the maximum value of 
+        ${maxIdleTime} milliseconds. It will be set to ${maxIdleTime} milliseconds instead.`
+      )
```
doom-weaver: i think we should remove logging like this.

Since this is a library we don't want to spam the user's context, and I don't see anywhere else in the driver we're doing this.

If we want to build in some sort of notification system we should probably think through a strategy before committing.

Let's make up for this gap by clarifying the documentation further.

```suggestion
```

doom-weaver:
Noting that @henryfauna mentioned @faunaee called out we should determine how this impacts streaming, as it could change our implementation. (good call out @faunaee !)

doom-weaver:
@henryfauna please prepare this as a new driver version following the guide here: https://faunadb.atlassian.net/wiki/spaces/DX/pages/3002335237/Drivers#JS-driver-release

README.md
```diff
@@ -296,8 +298,8 @@ const { Client, query: Q } = require('faunadb')
 async function main() {
   const client = new Client({
     secret: 'YOUR_FAUNADB_SECRET',
-    http2SessionIdleTime: Infinity,
-    //                    ^^^ Infinity or non-negative integer
+    http2SessionIdleTime: 1000,
+    //                    ^^^ Non-negative integer
```
faunaee:
```suggestion
    http2SessionIdleTime: 1000, // Must be a non-negative integer
```

doom-weaver (dismissed):

README.md
faunaee: Note that the comment on line 309 is no longer relevant, since `Infinity` is no longer being used.

src/Client.js
```diff
@@ -158,9 +158,10 @@ var values = require('./values')
  *   Sets the maximum amount of time (in milliseconds) for query execution on the server
  * @param {?number} options.http2SessionIdleTime
  *   Sets the maximum amount of time (in milliseconds) an HTTP2 session may live
- *   when there's no activity. Must either be a non-negative integer, or Infinity to allow the
- *   HTTP2 session to live indefinitely (use `Client#close` to manually terminate the client).
- *   Only applicable for NodeJS environment (when http2 module is used). Default is 500ms;
+ *   when there's no activity. Must be a non-negative integer, with a maximum value of 5000.
+ *   If an invalid value is passed a default of 500 ms will be applied. If a value
```
faunaee:
```suggestion
 *   If an invalid value is passed a default of 500 ms is applied. If a value 
```

test/client.test.js
```diff
@@ -412,6 +417,115 @@ describe('Client', () => {
...
+    var internalIdleTime
+    const maxIdleTime = 5000
+    const defaultIdleTime = 500
+
+    process.env.FAUNADB_HTTP2_SESSION_IDLE_TIME = '999'
+    client = util.getClient({
+      http2SessionIdleTime: 2500,
+    })
+    internalIdleTime = client._http._adapter._http2SessionIdleTime
+    expect(internalIdleTime).toBe(999)
```
faunaee, line 432: To me, it's weird that explicit client specification of `http2SessionIdeltTime` would be superceded by an environment variable. IMHO, the environment variable should be use to provide a customer-specified default.

test/stream.test.js
```diff
@@ -273,6 +273,48 @@ describe('StreamAPI', () => {
...
+      })
+
+      stream.start()
+
+      await util.delay(idleTime + 500)
+      assertActiveSessions(1)
+
+      const { ts: newTimestamp } = await client.query(q.Update(doc.ref, {}))
+      expect(newTimestamp).toBeGreaterThan(oldTimestamp)
+      await util.delay(idleTime + 1)
```
faunaee: Wouldn't it be worth checking that the stream is still open beyond the maximum value of `http2SessionIdleTime`? As written, this test doesn't really demonstrate that a stream is held open indefinitely, just that it is held open for 2002 ms (which is well within 5000ms).

README.md
```diff
@@ -281,13 +283,13 @@ have been resolved, the client will keep the session open for a period of time
 
 The `http2SessionIdleTime` parameter may be used to control how long the HTTP/2
 session remains open while the connection is idle. To save on the overhead of
```
faunaee:
```suggestion
session remains open while the query connection is idle. To save on the overhead of
```

README.md
```diff
@@ -281,13 +283,13 @@ have been resolved, the client will keep the session open for a period of time
...
-closing and re-opening the session, set `http2SessionIdleTime` to a longer time
---- or even `Infinity`, to keep the session alive indefinitely.
+closing and re-opening the session, set `http2SessionIdleTime` to a longer time.
+The default value is 500ms and the maximum value is 5000ms.
 
 While an HTTP/2 session is alive, the client will hold the Node.js event loop
```
faunaee:
```suggestion
While an HTTP/2 session is alive, the client holds the Node.js event loop
```

README.md
```diff
@@ -281,13 +283,13 @@ have been resolved, the client will keep the session open for a period of time
 
 The `http2SessionIdleTime` parameter may be used to control how long the HTTP/2
 session remains open while the connection is idle. To save on the overhead of
-closing and re-opening the session, set `http2SessionIdleTime` to a longer time
---- or even `Infinity`, to keep the session alive indefinitely.
+closing and re-opening the session, set `http2SessionIdleTime` to a longer time.
+The default value is 500ms and the maximum value is 5000ms.
```
faunaee, line 287:
```suggestion
The default value is 500ms and the maximum value is 5000ms.

Note that `http2SessionIdleTime` has no effect on a stream connection: a stream
is a long-lived connection that is intended to be held open indefinitely.
```

faunaee (changes requested):
I've made a few wording suggestions, wondered about the env var precedence, and asked about a change to the stream testing.

test/client.test.js
doom-weaver: this is the behavior today; so henry is maintaining the current behavior. I agree with you though that's its a bit odd.

https://github.com/fauna/faunadb-js/blob/v4/src/Client.js#L164-L165

https://github.com/fauna/faunadb-js/blob/v4/src/Client.js#L387

src/Client.js
```diff
@@ -384,17 +385,29 @@ Client.prototype._handleRequestResult = function(response, result, options) {
...
-  return {
-    shouldOverride: useEnvVar,
-    value: useEnvVar ? parsed : 500,
+  var value = defaultIdleTime
+  // attemp to set the idle time to the env value and then the configured value
+  const values = [envIdleTime, configuredIdleTime]
+  for (const rawValue of values) {
+    const parsedValue = parseInt(rawValue, 10)
+    const isNegative = parsedValue < 0
+    const isInfinity = rawValue === 'Infinity'
```
faunaee: Wouldn't it be easier to do this check first?

The overall check could be simpler, too:
```
// Handle Infinity, and NaN (for any other string)
const parsedValue = rawValue === 'Infinity'
  ? Number.MAX_SAFE_INTEGER
  : parseInt(rawValue, 10) || defaultIdleTime

// Handle upper bound
if (parsedValue > maxIdleTime) parsedValue = maxIdleTime

// Handle lower bound
if (parsedValue < 0) parsedValue = defaultIdleTime

return parsedValue
```

faunaee (commented):
I noticed that the code determine the value to use for `http2SessionIdleTime` seems more complex than it needs to be.

test/client.test.js
henryfauna: Yeah I definitely agree as well it is odd but changing that behavior here may be a bit out of scope

test/stream.test.js
henryfauna: The test should be updated now

faunaee (dismissed):
Thanks for applying my suggestions!

LGTM 👍

Note: I've updated the PR for DOCS-2096 to note that `http2SessionIdleTime` does not affect stream connections.

doom-weaver (dismissed):

doom-weaver (approved):

PR #653, opened 2022-09-07
[FE-2894] fix type definition for accessing cause of FaunaHttpErrorResponseContent

josStorer:
Fix type definition for accessing cause of FaunaHttpErrorResponseContent

### DIFF
--- a/src/types/errors.d.ts
+++ b/src/types/errors.d.ts
@@ -4,6 +4,7 @@ export type FaunaHttpErrorResponseContent = {
   errors: {
     code: string,
     description: string
+    cause: any;
   }[]
 }
 export module errors {
--- a/src/types/errors.d.ts
+++ b/src/types/errors.d.ts
@@ -2,9 +2,10 @@ import RequestResult from './RequestResult'
 
 export type FaunaHttpErrorResponseContent = {
   errors: {
+    position: (string | number)[] 
     code: string,
     description: string
-    cause: any;
+    cause?: FaunaHttpErrorResponseContent[];
   }[]
 }
 export module errors {


src/types/errors.d.ts
```diff
@@ -4,6 +4,7 @@ export type FaunaHttpErrorResponseContent = {
   errors: {
     code: string,
     description: string
+    cause: any;
```
ptpaterson, line 7: The cause field is an array of additional errors, recursively, so I think its type can be specified like this

```suggestion
    cause?: FaunaHttpErrorResponseContent[];
```

Here's an example raw response from a query where a UDF calls a UDF which then has an error:

```json
{
  "errors": [
    {
      "position": [  ],
      "code": "call error",
      "description": "Calling the function resulted in an error.",
      "cause": [
        {
          "position": [ "expr" ],
          "code": "call error",
          "description": "Calling the function resulted in an error.",
          "cause": [
            {
              "position": [ "expr", "concat", 0 ],
              "code": "invalid argument",
              "description": "String expected, Integer provided."
            },
            {
              "position": [ "expr", "concat", 1 ],
              "code": "invalid argument",
              "description": "String expected, Integer provided."
            }
          ]
        }
      ]
    }
  ]
}
```

Do you have any other examples?

ptpaterson: The cause field is an array of additional errors, recursively. Also, `position` should also be included, so I think its type can be specified like this

```suggestion
    position: (string | number)[] 
    code: string,
    description: string
    cause?: FaunaHttpErrorResponseContent[];
```

Here's an example raw response from a query where a UDF calls a UDF which then has an error:

```json
{
  "errors": [
    {
      "position": [  ],
      "code": "call error",
      "description": "Calling the function resulted in an error.",
      "cause": [
        {
          "position": [ "expr" ],
          "code": "call error",
          "description": "Calling the function resulted in an error.",
          "cause": [
            {
              "position": [ "expr", "concat", 0 ],
              "code": "invalid argument",
              "description": "String expected, Integer provided."
            },
            {
              "position": [ "expr", "concat", 1 ],
              "code": "invalid argument",
              "description": "String expected, Integer provided."
            }
          ]
        }
      ]
    }
  ]
}
```

Do you have any other examples?

ptpaterson:
Hi @josStorer thanks for the PR!

I submitted issue [FE-2894](https://faunadb.atlassian.net/browse/FE-2894) to get another pair of eyes on this.

src/types/errors.d.ts
josStorer: I think this is a complete enough example of an error response

It would be better to have a simple query example corresponding to the response, just as the documentation has done

So I think this example can replace the one in the current document, or add an extra example
https://docs.fauna.com/fauna/current/build/fql/errors?lang=javascript

Thank you all for your work

henryfauna (approved):

PR #654, opened 2022-09-12
Add optional, per-query parameters for providing a traceparent and tags

trevor-fauna:
### Notes
[FE-2382](https://faunadb.atlassian.net/browse/FE-2382)

Adds new, optional headers to driver, `traceparent` and `tags`, for providing a unique identifier per-query identifier, as well as KV-pair tags, to each submitted query. README was updated as well:

---------------
##### traceparent
A [W3C-compliant](https://w3c.github.io/trace-context) identifier for enabling distributed tracing across different
vendors. If not provided, one is automatically generated server-side and attached to the query. See
[Trace Context](https://w3c.github.io/trace-context) spec for more details.
```javascript
var data = client.query(q.Paginate(q.Collections()), {
  traceparent: "00-c91308c112be8448dd34dc6191567fa0-b7ad6b7169203331-01",
})
```

##### tags
Allows for associating user-provided tags with a query.
```javascript
var data = client.query(q.Paginate(q.Collections()), {
  tags: { key1: "value1", key2: "value2" },
})
```
----------------

### How to test
Test cases were added to assert the new fields. They can be executed via `yarn test`.


README.md
```diff
@@ -256,12 +259,31 @@ var helper = client.paginate(
...
+##### queryTimeout
 ```javascript
 var data = client.query(q.Paginate(q.Collections()), {
   queryTimeout: 100,
 })
 ```
 
+##### traceparent
+A [W3C-compliant](https://w3c.github.io/trace-context) identifier for enabling distributed tracing across different
+vendors. If not provided, one is automatically generated server-side and attached to the query. See
```
wildemat: Let customers know if they submit an invalid traceparent, they must override their provided one with the one returned in our response, since theirs would be silently dropped.

src/_http/index.js
```diff
@@ -131,6 +142,29 @@ HttpClient.prototype.execute = function(options) {
...
+  }
+  var result = ''
+  const entries = Object.entries(tags)
+  const entriesLength = entries.length
+  var index = 1
+  entries.forEach(([key, value]) => {
+    result += key + '=' + value
+    if (index++ < entriesLength) result += ','
+  })
+  return result
```
wildemat:
```javascript
return Object.entries(tags).map(e => e.join('=')).join()
```

maybe we want to validate the key:value pairs

silently throw out bad ones
```javascript
let invalidTag = (value => [undefined, "", null].includes(value))
return Object.entries(tags)
  .filter(e => !e.some(invalidTag))
  .map(e => e.join('=')).join() 
```

Exception
```javascript
let invalidTag = (value => [undefined, "", null].includes(value))
return Object.entries(tags)
  .map(e => {
    if (e.some(invalidTag)){
      throw new Error(`Invalid tag: ${e}`)
    }
    return e.join('=')
  }).join() 
```

faunaee: Missing ',' from the final join. Seems like it should be:

```javascript
return Object.entries(tags).map(e => e.join('=')).join(',')
```

wildemat: comma is the default, but yes better to be verbose

trevor-fauna:
```javascript
return Object.entries(tags).map(e => e.join('=')).join(',')
```

Much cleaner, thanks.

>maybe we want to validate the key:value pairs
    silently throw out bad ones

Exception is preferred to silently discarding. The only reason we silently discard invalid traceparents is because the W3C spec defines it that way, but in general, we should try to consider customer requests immutable. That said, let me sync with product on a reasonable length of tags and update that in a separate PR.

README.md
trevor-fauna: Updated

wildemat (approved):

src/_http/index.js
```diff
@@ -131,6 +142,23 @@ HttpClient.prototype.execute = function(options) {
...
+}
+
+function parseTags(tags) {
+  if (tags === undefined || tags == null || tags == '') return null
+  if (typeof tags != 'object') {
+    throw new Error('Tags must be provided as an object!')
+  }
+  return Object.entries(tags)
+    .map(e => e.join('='))
+    .join(',')
```
doom-weaver, line 159: trim the keys and values?

reject empty/blank keys?

example code to help clarify:

```
 > x = {'': 'dog'}
{ '': 'dog' }
 > ' cat '.trim()
'cat'
```

trevor-fauna: Yeah, gonna add validation in a follow-up (including max key/value length, etc.).

PR #656, opened 2022-09-15
Test that endpoint param can contain trailing slashes

ptpaterson:
## Problem

The new endpoint param introduced in #652 should handle trailing slashes (`https://db.fauna.com/` for example).

## Solution

The HTTP adapters already handle this.  We want to make sure we test for it, though.

## Testing

Test ping on clients created using endpoint parameters with trailing slashes.

### DIFF
--- a/test/client.test.js
+++ b/test/client.test.js
@@ -44,6 +44,23 @@ describe('Client', () => {
     })
   })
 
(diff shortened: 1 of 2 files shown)


faunaee (approved):
I only performed a visual inspection, but it _looks_ good. 👍

PR #658, opened 2022-09-15
Tag validation

trevor-fauna:
### Notes
[FE-2693](https://faunadb.atlassian.net/browse/FE-2693)

Note: This will be published to npm as a distinct tag pointed at our incremented version, `4.8.0-beta.1`. This will allow beta customers to download the specific driver version while still allowing us to merge the code back on `v4` afterwards.

### How to test
`yarn test`

### DIFF
--- a/src/_http/index.js
+++ b/src/_http/index.js
@@ -151,12 +151,52 @@ function isValidTraceparentHeader(traceparentHeader) {
 
 function parseTags(tags) {
   if (tags === undefined || tags == null || tags == '') return null
-  if (typeof tags != 'object') {
-    throw new Error('Tags must be provided as an object!')
-  }
+  validateTags(tags)
   return Object.entries(tags)
     .map(e => e.join('='))
     .join(',')
+}
+
+function validateTags(tags) {
+  const maxTagEntries = 25
+  if (typeof tags != 'object') {
+    throw new Error('Tags must be provided as an object!')
+  }
+  if (Object.entries(tags).length > maxTagEntries) {
+    throw new Error('Tags are limited to 25 entries')
+  }
+  validateTagKeys(Object.keys(tags))
+  validateTagValues(Object.values(tags))
+}
+
+function validateTagKeys(keys) {
+  const maxKeyLength = 40
+  keys.forEach(key => {
+    if (typeof key != 'string') {
+      throw new Error(`Provided key, ${key}, must be a string`)
+    } else if (key.length > maxKeyLength) {
+      throw new Error(
+        `Provided key, ${key}, must not exceed maximum length of ${maxKeyLength} characters`
+      )
+    } else if (!/^\w+$/.test(key)) {
+      throw new Error(`Provided key, ${key}, contains invalid characters`)
+    }
+  })
+}
+
+function validateTagValues(values) {
+  const maxValueLength = 80
+  values.forEach(value => {
+    if (typeof value != 'string') {
+      throw new Error(`Provided value, ${value}, must be a string`)
+    } else if (value.length > maxValueLength) {
+      throw new Error(
+        `Provided value, ${value}, must not exceed maximum length of ${maxValueLength} characters`
+      )
+    } else if (!/^\w+$/.test(value)) {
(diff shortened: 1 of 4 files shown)


src/_http/index.js
```diff
@@ -151,14 +151,54 @@ function isValidTraceparentHeader(traceparentHeader) {
...
+    throw new Error('Tags must be provided as an object!')
+  }
+  if (Object.entries(tags).length > maxTagEntries) {
+    throw new Error('Tags are limited to 25 entries')
+  }
+  validateTagKeys(Object.keys(tags))
+  validateTagValues(Object.values(tags))
+}
+
+function validateTagKeys(keys) {
```
doom-weaver, line 172: feels a bit aggressive to fail the request due to problems with tags.

I recall you thought about the tradeoffs between cleaning up the data vs blowing up - but could we discuss that again here?

Another option would be to force values to a string + trim up the input; and clean out tags that don't meet the spec; and document that as the contract.

I'm imaging a scenario where a customer has automation injecting tags - not sure they'd want to start failing their queries due to issues with the tag formats.

Speaking for myself as a consumer; i'd prefer to carry on with my business logic and spot the tagging issue offline when i notice my observability platform doesn't match what i expect.

trevor-fauna: TL;DR - Fail fast. If I'm doing something wrong, tell me, and tell me right away - because as my system gets bigger, it becomes more difficult to track down the problem.

Longer story is we ended up discussing this a bit on the DB side as well, where we have similar validation going into place. It boils down to whether you consider tags as part of the "request" (more specifically, the query request, and thus should throw 4XX) or if it's simply metadata that can be ok to fail (with the query returning 200). Whatever that decision, the client-side validation should match with the intent that we simply perform the validation as early as possible. When you consider the varied things that customers tend to use tags for - one of the most critical is often billing/analytics use cases. It's often not acceptable to let things go unnoticed there for long periods of time, and if you do, it becomes really painful to fix in retrospect (as most product devs know far too well...😒). I wouldn't want to just add this validation to a new use case/feature. However, in this case, we're introducing it, as well as these semantics, out of the gate. If you do something wonky, we'll let you know, and you can fix it. Perhaps, without knowing what you're using the tags for, and assuming that it's fairly orthogonal to your use case...then sure, we can silently fail/sanitize. However, if you assume that tagging becomes a mission-critical part of a customer's business, then it actually makes sense to fail immediately. They'll be able to fix the issue faster, and then move on with debugging the rest of their data flow.

As far as sanitizing the input - I think it's dangerous to start doing that on a customer request as becomes more difficult, as a consumer, to detect where things fail. If my key is too long, do you just blindly truncate it? Do you truncate the head, or tail? Is that configurable? One gap right now where we take the approach you're suggesting is on duplicate keys, which Javascript will just silently overwrite. I considered that an acceptable gap for most developers given that it's a language-level semantic, but I'd also prefer to throw an exception even for this case for the above reasons on how tags are typically used.

doom-weaver: so the database is going to throw 400s for bad tags?

Is the same true for bad trace ids?

trevor-fauna: For bad tags, we will yeah: https://faunadb.atlassian.net/browse/ENG-3762

TraceIds are more interesting. Right now, we're generating a new one and returning that in the response if the one in the request was invalid. We're doing that because of the following verbiage in the spec:

> If the vendor supports the version number, it validates trace-id and parent-id. If either trace-id, parent-id or trace-flags are invalid, the vendor creates a new traceparent header and deletes tracestate.

Generally, there's only one version number in the spec right now, which is `00`, so...yes, we support that. However, the spec is less clear when the traceId is completely mangled, like `#@#$@#...ThisIsNotARealTrace`:

> Vendors MUST ignore the traceparent when the parent-id is invalid (for example, if it contains non-lowercase hex characters).

For now, we're just generating a new one as the default - but, it's certainly worth digging into the spec further once more people have had a chance to read through and see the functionality in action.

doom-weaver (approved):

PR #659, opened 2022-09-16
[FE-2693] Precompile tag validation regex

trevor-fauna:
### Notes
[FE-2693](https://faunadb.atlassian.net/browse/FE-2693)

Addressing out-of-band feedback on https://github.com/fauna/faunadb-js/pull/658

### How to test
`yarn test`

### DIFF
--- a/src/_http/index.js
+++ b/src/_http/index.js
@@ -169,6 +169,8 @@ function validateTags(tags) {
   validateTagValues(Object.values(tags))
 }
 
+const AlphaNumPattern = new RegExp(/^\w+$/)
+
 function validateTagKeys(keys) {
   const maxKeyLength = 40
   keys.forEach(key => {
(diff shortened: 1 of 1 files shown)


src/_http/index.js
```diff
@@ -169,6 +169,8 @@ function validateTags(tags) {
   validateTagValues(Object.values(tags))
 }
 
+const AlphaNumPattern = new RegExp(/^\w+$/)
```
doom-weaver, line 172: try putting this at line 6 and i think it should still work mint it just once for any call to HttpClient.

trevor-fauna: Yeah, I wasn't sure - felt weird there as a convention with all the imports. Should I make it all caps?

doom-weaver: yeah that sounds like a good call.

doom-weaver: you got another regex at line 147 btw.

doom-weaver: actually reading a bit more: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions

it sounds like a literal will cause it to be compiled once.

So we may be good.

PR #661, opened 2022-09-29
Fix tags type

doom-weaver:
### Notes
Tags type only allows a single item, literally keyed "string", in the `{ "string": <value of tag>: string}` type definition.

Fix this so that it allows any number of string typed keys with string typed values

### DIFF
--- a/src/types/Client.d.ts
+++ b/src/types/Client.d.ts
@@ -37,7 +37,7 @@ export interface QueryOptions
   > {
   signal?: AbortSignal
   traceparent?: string
-  tags?: { string: string }
+  tags?: { [key: string]: string }
 }
 
 type StreamFn<T> = (
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "faunadb",
-  "version": "4.8.0-beta.1",
+  "version": "4.8.0-beta.2",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "faunadb",
-      "version": "4.8.0-beta.1",
+      "version": "4.8.0-beta.2",
       "hasInstallScript": true,
       "license": "MPL-2.0",
       "dependencies": {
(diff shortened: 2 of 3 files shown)


adambollen:
Looks like `package-lock.json` has a ton of changes - is that expected?

doom-weaver:
> Looks like `package-lock.json` has a ton of changes - is that expected?

All i did is run npm install. 

I think what happened was this branch in the package.lock refers to a different version of the driver:

https://github.com/fauna/faunadb-js/blob/fauna_logs/package-lock.json#L3

compare:

https://github.com/fauna/faunadb-js/blob/3468f6e7a86d77284e54eff8cebea09d0d531b91/package-lock.json#L3

adambollen:
> I think what happened was this branch in the package.lock refers to a different version of the driver:
> 
> https://github.com/fauna/faunadb-js/blob/fauna_logs/package-lock.json#L3
> 
> compare:
> 
> https://github.com/fauna/faunadb-js/blob/3468f6e7a86d77284e54eff8cebea09d0d531b91/package-lock.json#L3

Oh, that makes sense.  We'll need to bump that to `-beta.2` before releasing also.

src/types/Client.d.ts
```diff
@@ -37,7 +37,7 @@ export interface QueryOptions
   > {
   signal?: AbortSignal
   traceparent?: string
-  tags?: { string: string }
+  tags?: { [key: string]: string }
```
adambollen, line 40: I'm a little confused; when I was running queries with multiple tags, they were showing up correctly in the logs.  Shouldn't I expect that not to work?

doom-weaver: did you use typescript or raw javascript?

If you use types it would refuse to compile.

adambollen: Oh right, that's the ticket.  Thanks.

doom-weaver: > Oh, that makes sense. We'll need to bump that to `-beta.2` before releasing also.

Sure I can do that now.

adambollen (approved):

PR #676, opened 2023-02-28
Update Expr.d.ts

william1616:
### Notes
This is a minor proposed change to correct the type definition of the Expr class - Expr.d.ts was missing the `toJSON` function so in typescript environments it wasn't possible to call Expr.toJSON() 

Example below using faunadb-js@4.8.0
image

### DIFF
--- a/src/types/Expr.d.ts
+++ b/src/types/Expr.d.ts
@@ -3,5 +3,6 @@ export default class Expr {
 
   readonly _isFaunaExpr?: boolean
   toFQL(): string
+  toJSON(): string
   static toString(expr: Expr): string
 }


src/types/Expr.d.ts
```diff
@@ -3,5 +3,6 @@ export default class Expr {
 
   readonly _isFaunaExpr?: boolean
   toFQL(): string
+  toJSON(): string
```
ptpaterson, line 6:
```suggestion
  toJSON(): Record<string, any>
```

ptpaterson (changes requested):

src/types/Expr.d.ts
ptpaterson: The `toJSON()` metho returns the underlying object. In that, "JSON" is a little misleading since it's not a string. JS makes it trivial to convert an object to a JSON string, so it's (too) easy to conflate the two.

ptpaterson:
Hi @william1616 and thanks for the submission!

You've been working with my colleague, Zee, and I gave him bad advice. I am sorry for the confusion, but `Expr.toJSON()` is not necessary for you to call, so the method does not need to be public.

In order to get the wire-protocol representation of the FQL query, you may execute `JSON.stringify(x)`.

This works because all FQL functions are `Expr` instances and when you use `JSON.stringify` it picks up the custom `toJSON` function.

Please let us know if that solves the problem. If so, we can close this PR. If not, we'll keep working on it.

Thanks!

william1616:
thanks @ptpaterson (just replied with you on cc to Zee's email) - i'll close this PR

